PR #13, opened 2024-06-05
Update README.md

gkhngyk:
- Fix parameters redirect problem


unclecode:
I already accepted the other request, both are the same, so I close this  one.

Issue #20, opened 2024-06-14
result.markdown is missing words when there's a span enclosed in a h1 tag in the crawled page


aravindkarnam:
Sometimes a span is introduced in headings to either change text color, bold or italicize certain words in a heading or paragraph etc. Such spans enclosed in HTML text/heading tags seem to get left out in `result.markdown`.

Consider the following pages:
1. https://urbanpiper.com, inside h1 tag `<h1>"Manage your food delivery channels on <span class="purple-text">one dashboard</span></h1>"`, when this page is crawled and result is produced, in `result.markdown` "one dashboard" is missing, leading to incomplete sentence.
2. https://reelsify.app, inside h1 tag `<h1>"Engage your customers with short <span>snackable</span> videos"</h1>`, this same bug is reproducible.(the word "snackable" is missing in result.markdown)

This seems like a bug.  Please let me know if you need any help in reproducing this bug.

unclecode:
@aravindkarnam Very interesting! Then I'll flatten those header tags that contain inner tags. Spans or similar tags within headers don't add extra value. A header tag already signifies importance, so emphasizing parts within it is redundant. I'll convert something like `<h1>Manage your food delivery channels on <span class="purple-text">one dashboard</span></h1>` to `<h1>Manage your food delivery channels on one dashboard</h1>`.

I'll update it soon. Thanks for sharing.

aravindkarnam:
@unclecode Raised a PR for this. Please review.

unclecode:
@aravindkarnam All done, and now you can check the utils.py. Again, appreciate your contribution. Please stay connected and share any other issues you face with Crawl4ai. I close this PR.

PR #24, opened 2024-06-17
Remove inline styling tags and merge their contents into their parent elements

aravindkarnam:
Ref: https://github.com/unclecode/crawl4ai/issues/20

It's a common practice to add inline style tags to text to bold, italicize or change colour of certain words in a paragraph to emphasise them to human readers of a site. However, this is not of any relevance to LLMs, since the words are already in their context.

In crawl4ai `remove_empty_and_low_word_count_elements` is removing these tags altogether and some valuable data is being lost. Therefore we should merge the contents of these tags into their parent elements before this step.

### DIFF
--- a/crawl4ai/utils.py
+++ b/crawl4ai/utils.py
@@ -250,6 +250,15 @@ def get_content_of_website(url, html, word_count_threshold = MIN_WORD_THRESHOLD,
         # Replace all "pre" tags with their inner text
         body = replace_pre_tags_with_text(body)
 
+
+        # List of inline styling tags
+        inline_styling_tags = ['b', 'i', 'u', 'span', 'del', 'ins', 'sub', 'sup', 'strong', 'em', 'code', 'kbd', 'var', 's', 'q', 'abbr', 'cite', 'dfn', 'time', 'small', 'mark']
+        # Iterate over each tag and unwrap it
+        for tag in inline_styling_tags:
+            for match in soup.find_all(tag):
+                match.unwrap()
+                
+
         # Recursively remove empty elements, their parent elements, and elements with word count below threshold
         def remove_empty_and_low_word_count_elements(node, word_count_threshold):
             for child in node.contents:


unclecode (commented):
I like it, but I have one concern. Instead of just unwrapping them, we should maintain part of their semantics. One way is to format them in markdown, so something like `<h1>Hello <i>World</i></h1>` would be `<h1>Hello *World*</h1>`. This approach retains more semantics. Tags like "code", "abbr", and "cite" still matter, so we should either use markdown or a custom formatter function to convert these tags in our "LLM-friendly" way. For example, `<abbr title="World Health Organization">WHO</abbr>` would become `WHO (World Health Organization)`, as even markdown formatters might miss this. My preferences is:

- `<b>bold text</b>` can be `**bold text**`
- `<i>italic text</i>` can be `*italic text*`
- `<u>underlined text</u>` can be `__underlined text__`
- `<span>span text</span>` can be `span text`
- `<del>deleted text</del>` can be `~~deleted text~~`
- `<ins>inserted text</ins>` can be `++inserted text++`
- `<sub>subscript text</sub>` can be `~subscript text~`
- `<sup>superscript text</sup>` can be `^^superscript text^^`
- `<strong>strong text</strong>` can be `**strong text**`
- `<em>emphasized text</em>` can be `*emphasized text*`
- `<code>code text</code>` can be `` `code text` ``
- `<kbd>keyboard text</kbd>` can be `` `keyboard text` ``
- `<var>variable text</var>` can be `_variable text_`
- `<s>strikethrough text</s>` can be `~~strikethrough text~~`
- `<q>quoted text</q>` can be `"quoted text"`
- `<abbr title="Title Text">abbr</abbr>` can be `abbr (Title Text)`
- `<cite>citation</cite>` can be `_citation_`
- `<dfn>definition</dfn>` can be `_definition_`
- `<time datetime="2024-06-17">time text</time>` can be `time text (dated: 2024-06-17)`
- `<small>small text</small>` can be `(small text)`
- `<mark>highlighted text</mark>` can be `==highlighted text==`

It would be great if you're interested in handling this. Could you create the function and add it to line 252?

aravindkarnam:
@unclecode 
> I have one concern. Instead of just unwrapping them, we should maintain part of their semantics. One way is to format them in markdown, so something like `<h1>Hello <i>World</i></h1> would be <h1>Hello *World*</h1>`

So in my case I intend to crawl a site, make vector embeddings and load it into a vector DB. So that when a specific context is required later, LLM can fetch it from the vector database. So, what impact would markdown semantics have on vector embeddings of corresponding words?

For example, if the crawled text is `<span> our <b>support email</b> is support@acmecorp </span>` and we get that converted into `our *support email* is support@acmecorp`. Later when a user asks the question, "What is email address of acme corp's support" would the similarity search know/get that `support email` and `*support email*` are the same thing meaning-wise? Also what benefit would LLM get from word level Markdown semantics? 

Headings etc in markdown would give an overview of the information hierarchy of the document, so it makes sense to store them and even pass them to LLM. Unless we have a lot of benefits from word-level markdown semantics, we can conserve cost-incurred by reducing no. of tokens. (both in getting vector embeddings and context window length that gets used up while invoking LLM)

Sorry if this line of thought comes across as a noob, I've just started learning about building apps with LLM. Please let me know what you think.

unclecode:
@aravindkarnam No worries, I really appreciate you sharing your thoughts and engaging in this discussion.

1. I will add the ability I mentioned earlier as an option, allowing users to choose between pure text, as you suggested, or text represented with additional symbols, as I proposed.
   
2. It's important to note that tags like "abbr", "dfn", and "mark" carry significant semantic meaning. A well-designed vector database, especially those utilizing large embedding models trained on such nuances, can leverage this information. While the impact of a "<b>" tag may not be as clear, the importance of an "abbr" tag is quite evident.

3. A key use case for my library is training embedding models, which I have personally utilized. In my experience, I built a custom tokenizer sensitive to such tags. When we trained the embedding model with this approach, it outperformed many current models in the top 10 leaderboard benchmarks, demonstrating the value of preserving these nuances.

4. I will integrate this functionality into the library and add a flag to allow users to decide whether to replace inline tags with their HTML or convert them based on the conversion map I explained.

5. If you fine, I plan to pull this PR into a new brach, there I modified the codebase and merge it to the main request.

aravindkarnam:
@unclecode An option/flag to choose b//n pure text vs markdown semantics sounds perfect. That way users will have a choice to optimise/prioritize for whichever option works best for them.

Please proceed with PR as you suggested.

unclecode:
@aravindkarnam Thank you for the help on this PR! I've merged your changes into a new branch called `format-inline-tags` for further modifications before merging into the main branch. I will release it soon with the new version. This will ensure we incorporate all necessary enhancements. Your contribution is much appreciated!

Issue #26, opened 2024-06-20
Add some textual context to images, videos and audio in `result.media`, so that it can be useful in RAG


aravindkarnam:
## Issue
Currently, all images, video and audio in the crawled page are returned in `result.media` like follows.  
```
{
audio:[],
images:[
{
src:"https://cdn.com/path/to/image",
alt:"mobile_integrations_560px",
type:"image"
},
.
.
.
],
videos:[]
}
```

Each entry in this object list contains a link/url from where the asset could be downloaded, the text passed in the `alt` attribute of the corresponding `<img>` HTML tag and a `type` field(since the entries are already grouped by audio, video and images not sure why we need the type field again)

Other than this there's no useful information regarding what may be the content of the media file (It could be architecture diagram, testimonial, partners, integrations etc). 

So we have to entirely rely on the `alt` attribute to vectorise and store against it the link to the media file so that it can be retrieved during RAG. Also theres many times where I found utterly garbage if not misleading values in alt attribute of image tags of even well known companies. 

I'm talking about a use case like perplexity (but for a company internal knowledge base etc)

![Screenshot 2024-06-20 at 4 09 55 PM](https://github.com/unclecode/crawl4ai/assets/15815124/9a22f1b5-e0dc-48bf-b922-8929a63cef50)

## Suggestion

I suggest that we add some textual context based on the position ie where the image is found with in the page. Perhaps the closest paragraph or header (right above the image). @unclecode please let me know your thoughts on this or if there's a better way to do RAG on images, for user queries.

unclecode:
Very interesting point that you mentioned, and it's actually been on my mind. Initially, I started by just extracting the media tags to have them ready. I have two strategies in mind. 

First, I'm looking for a very small image-to-text model to annotate images and create descriptions. I've found a few or am working on one myself. This model needs to be efficient, not large, as we only need small captions and want to run it in parallel without heavy GPU dependence. There are many media sources on a page and need to be fast, otherwise it will be time consuming.

Second, there's a statistical NLP approach uses TF-IDF to gather information around the image, pulling in relevant text from the website, maybe even from other parts that refer to the image initially. This doesn't need a model, just a statistical method.

Since you brought this up, I'll start with the statistical approach while continuing my research and development on the LLM. Thank you for bringing this to my attention.

aravindkarnam:
[...]

PR #60, opened 2024-07-21
Filter images from HTML and extract useful text from nearest parent component

aravindkarnam:
@unclecode Here's the PR for enhancement #26 

- I've added the code under the newly implemented `get_content_of_website_optimized` insted of the `get_content_of_website` method, since the `run` function calls the former now.

- All helper methods that are entirely used only within another function then I've defined it under the block scope of calling function.

- The core functionality is with in the `process_image` function in `utils.py`. I've noticed that you have implemented the `process_element` function to be recursive and left a stub for image handling there(to add images to `result.media['images']`). 
However the `process_image` depends on the whole array of images (it scores images based on their position in the array), so I feel it would be cumbersome to implement it recursively. Also there is no nesting in images, so you wouldn't get any added advantage also by calling `process_image` in recursion.
So I've deleted the stub in the recursive `process_element` function, instead extracted a flat array of images and looped over with `process_image` function, to add qualifying images to `result.media[`images`]` along with the extracted text and score.

- I moved the cutoff threshold to `config.py` under the name `IMAGE_SCORE_THRESHOLD` as constant. So users can tweak around withit based on their usecase.

I tested this locally and it's working as expected. Please add comments on the PR, for any changes/bugs, I'll address them. I learned a lot working on this enhancement, thanks for the opportunity!

PS: There was an import error in `model_loader.py`, I fixed it so that code can run. It's not relevant to the enhancement, so calling it out here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [v0.2.75] - 2024-07-19
+
+Minor improvements for a more maintainable codebase:
+
+- 🔄 Fixed typos in `chunking_strategy.py` and `crawler_strategy.py` to improve code readability
+- 🔄 Removed `.test_pads/` directory from `.gitignore` to keep our repository clean and organized
+
+These changes may seem small, but they contribute to a more stable and sustainable codebase. By fixing typos and updating our `.gitignore` settings, we're ensuring that our code is easier to maintain and scale in the long run.
+
 ## [v0.2.74] - 2024-07-08
 A slew of exciting updates to improve the crawler's stability and robustness! 🎉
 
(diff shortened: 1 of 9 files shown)


unclecode:
@aravindkarnam I tested, all doing well. I just made two small changes. 

- First, I noticed that sending a HEAD request to get the image size was really slowing things down, so I commented that part out for now until we find a better solution.

- Second, in your scoring, I noticed an issue with the image format splitting. When splitting the text, there are situations where it returns the format with a dot in it, especially if the file name already has multiple dots. I trimmed the dot, and now everything works fine. 

That's all. and it's really amazing. Thx for the contribution. I've incorporated your changes into the upcoming version branch (main-75) rather than merging directly into main. This allows me to include your fix in our next release while keeping our current version stable.

PR #64, opened 2024-07-27
Update LocalSeleniumCrawlerStrategy to resolve ChromeDriver version m…

ketonkss4:
…ismatch issue

This resolves the following error: `selenium.common.exceptions.SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 114` 

Windows users are getting.


unclecode:
@ketonkss4 Very interesting I wasn't aware this. Anyway thank you for your contribution! We've incorporated your changes into our upcoming version branch (main-75) rather than merging directly into main. This allows us to include your fix in our next release while keeping our current version stable.

PR #65, opened 2024-07-31
Fix dockerfile + docker-compose

FractalMind:
Fix dockerfile missing slowapi package + make .env able to overwrite the default number of workers in docker-compose.yml


FractalMind:
Duplicated

Issue #71, opened 2024-08-07
Support extraction of media files from <source> tag in HTML


aravindkarnam:
[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source) tags pointing multiple media file formats, are used inside other media tags like audio and video. eg:
```
<video controls>
  <source src="foo.webm" type="video/webm" />
  <source src="foo.ogg" type="video/ogg" />
  <source src="foo.mov" type="video/quicktime" />
  I'm sorry; your browser doesn't support HTML video.
</video>
```
This way browsers can pick the first format they support. Currently we are only extracting based on src attribute on tags like audio, video etc. 

For example the following [site](https://meetmadison.ai/) has several high quality videos inside, but these videos are not getting extracted.

@unclecode I'll try to work on it this weekend. Will raise a PR once completed.

unclecode:
@aravindkarnam Oh you’re absolutely right—great job noticing that! Please go ahead and let me know the results. Also, consider what other similar tags we might be missing. For example:

```
<picture>
  <source media="(min-width:650px)" srcset="img_pink_flowers.jpg">
  <source media="(min-width:465px)" srcset="img_white_flower.jpg">
  <img src="img_orange_flowers.jpg" alt="Flowers" style="width:auto;">
</picture>
```

unclecode:
@aravindkarnam Hi, I hope you are doing well. Did you find time to work on this?

aravindkarnam:
@unclecode Hey! Doing well. I've added support for extraction of media from source tags in both Audio and Video tags. Added a commit [here](https://github.com/aravindkarnam/crawl4ai/commit/9ed155112516b2b2da60970303972adca3e3ac29)

As for the picture tag, I've read in the docs that
> The <img> element is <ins>required</ins> as the last child of the `<picture>` element, as a fallback option if none of the source tags matches.

Since img tag is required child of `<picture>` tag, the existing code for image extraction will catch it. Hence did not add any separate handling for it.

While I was at it, I've extended the extracted of text description from nearest parent element to video and audio also. The results looks splendid. I believe we don't need any filtering for videos and audio similar to images, because there aren't usually as many in a website unless it's a video site like youtube etc.

Please create a staging branch for next release, I can send a PR to it.

unclecode:
[...]

PR #75, opened 2024-08-19
Added support to source tags wrapped inside video and audio tags. Ext…

aravindkarnam:
1. Added support for extracting videos and audio from source tags. Picture tag will have one required  child, hence existing code for image extraction will handle it.
2. Extended the text description extraction to video and audio elements in media. 

https://github.com/unclecode/crawl4ai/issues/71


unclecode:
Thx @aravindkarnam It looks good to me, I merge it to stage an will do more tests.

Issue #77, opened 2024-08-26
Langchain integration for crawl4ai [ Request for comments]


aravindkarnam:
## Requirements

1. Crawl4ai's output should be encapsulated in the `Document`( with attributes `page_content: str` & `metadata: dict`) [object](https://api.python.langchain.com/en/latest/documents/langchain_core.documents.base.Document.html) of a `Crawl4aiLoader` that extends a [langchain Document Loader](https://api.python.langchain.com/en/latest/document_loaders/langchain_core.document_loaders.base.BaseLoader.html#langchain_core.document_loaders.base.BaseLoader), so that it can be fed to a chain for further processing.
2. Instantiation of the `crawl4ai document loader` should support the complete interface of crawl4ai's `WebCrawler` class. Similarly the `load` method should support the complete interface of the `run` method of crawl4ai.

## References
- https://python.langchain.com/v0.1/docs/contributing/integrations/
- https://python.langchain.com/v0.2/docs/how_to/document_loader_custom/#interface
- [Baseloader](https://api.python.langchain.com/en/latest/document_loaders/langchain_core.document_loaders.base.BaseLoader.html#langchain_core.document_loaders.base.BaseLoader)
- [Document](https://api.python.langchain.com/en/latest/documents/langchain_core.documents.base.Document.html#)

## Specs

- [ ] Eventually we would want to make this integration available in `langchain-community` [package](https://github.com/langchain-ai/langchain/tree/master/libs/community/langchain_community/document_loaders). However we would need some trial and errors + testing before raising a PR directly to that repo. For the first-cut of Langchain integration, we have **two options**.
1.  Fork a new repository from [langchain community repo](https://github.com/langchain-ai/) and start implementation in the `/libs/community/langchain_community/document_loaders`. Build a couple of versions, make it available to community and finally when we are ready with integration tests, units tests and documentation with a mature first version, we can raise a PR to the main langchain repo. [Langchain's community contribution guidelines](https://python.langchain.com/v0.1/docs/contributing/integrations/) suggests this approach.
[...]

PR #85, opened 2024-09-04
Langchain - Document loader basic functionality i.e `lazy load` implemented Issue #77

aravindkarnam:


rajesh-mdg (approved):

Huathi (approved):

Gulamhusin (approved):

simjak:
@aravindkarnam, could we have LangChain dependencies as optional extras? LangChain is quite a complex and heavy library, so it would be better to have a core crawl service and then add extra dependencies if needed, rather than binding them from the start.

unclecode:
@aravindkarnam Proceed, it totally make sense

PR #93, opened 2024-09-12
Add support for non-ASCII characters in JSON output

rangehow:
## Description
This PR modifies the JSON dumping process to support non-ASCII characters, enhancing support for multiple languages.

## Changes
- Modified the `json.dumps()` call to include `ensure_ascii=False`

## Motivation
The current implementation limits the output to ASCII characters, which can cause issues with non-Latin scripts and special characters. This change allows for proper representation of a wider range of languages and characters in the JSON output.

## CASE
before：
image
after：
image

### DIFF
--- a/crawl4ai/web_crawler.py
+++ b/crawl4ai/web_crawler.py
@@ -225,7 +225,7 @@ class WebCrawler:
 
                 sections = chunking_strategy.chunk(markdown)
                 extracted_content = extraction_strategy.run(url, sections)
-                extracted_content = json.dumps(extracted_content, indent=4, default=str)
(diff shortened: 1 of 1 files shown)


unclecode:
@rangehow I appreciate your suggestion. Actually, in the new version, we've applied this to wherever we dump JSON. We're planning to release this new version today or in a few days, and it's worth noting that you'll be using Playwright, which supports 100% asynchronous operations and includes several new, cool features. I'll add you to the list for Contributors, and we'll apply this to the upcoming release. Thank you again. I'm happy to close and accept your pull request, but I'll do it directly on the main branch. Thank you so much.

PR #108, opened 2024-09-29
Adding a parameter to disable SSL verification during crawling

ifaddict1:
Dears, while building a tool with your wonderful library, I realized there was no support for disabling SSL verification for websites that either use HTTP only or use HTTPS with an invalid certificate. So I added this option in the "AsyncPlaywrightCrawlerStrategy" class. I think this change could benefit a lot of people and prevent breaking certain scripts. I also added a test. Let me know if I should modify anything


aravindkarnam:
@ifaddict1 Thanks for raising this wonderful PR. I can see there are now many merge conflicts and the library has also evolved quite a lot since you raised this PR. But this(bypassing SSL verification) is a really important ask.

Going forward please raise a Feature request in the discussions. I already added this [one](https://github.com/unclecode/crawl4ai/discussions/525) for you.  

From feature requests we can properly plan, develop and release with full visibility to the community. Let us consider this feature and get back to you.

Cc: @unclecode

Issue #116, opened 2024-10-01
Using Proxy


shing-li:
There's a example code for proxy, but I didn't see the proxy parameter in class AsyncWebCrawler()

unclecode:
Hi, This is code sample to try proxy:

```
async def use_proxy():
    print("\n--- Using a Proxy ---")
    print(
        "Note: Replace 'http://your-proxy-url:port' with a working proxy to run this example."
    )
    # Uncomment and modify the following lines to use a proxy
    async with AsyncWebCrawler(verbose=True, proxy="http://your-proxy-url:port") as crawler:
        result = await crawler.arun(
            url="https://www.nbcnews.com/business",
            bypass_cache=True
        )
        print(result.markdown[:500]) 
```

Feel free to let me know if you need any thing :)

PratikHotchandani22:
Hi, I am getting this error after running the above code:

AsyncWebCrawler.__init__() got an unexpected keyword argument 'proxy'

anthonytison:
> Hi @unclecode, I am getting this error after running the above code:
> 
> AsyncWebCrawler.**init**() got an unexpected keyword argument 'proxy'

I got the same issue :(

SwallowData:
TypeError: AsyncWebCrawler.__init__() got an unexpected keyword argument 'proxy'
me too

aravindkarnam:
Here's the documentation on using [proxy](https://crawl4ai.com/mkdocs/advanced/proxy-security/)

PR #119, opened 2024-10-02
Add proxy functionality to AsyncWebCrawler and AsyncPlaywrightCrawler

neelthepatel8:
This PR adds proxy functionality to the AsyncWebCrawler and AsyncPlaywrightCrawlerStrategy classes.

Linked Issue: #116 

- Modified AsyncWebCrawler to accept a `proxy` parameter.
- Updated AsyncPlaywrightCrawlerStrategy to handle proxy settings when launching the browser.
- Added usage example in the documentation.

This allows users to specify a proxy server for crawling, which can be useful for accessing content behind firewalls.

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -59,6 +59,7 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             'after_goto': None,
             'before_return_html': None
         }
+        self.proxy = kwargs.get("proxy")
 
     async def __aenter__(self):
         await self.start()
@@ -73,7 +74,6 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
         if self.browser is None:
             browser_args = {
                 "headless": self.headless,
-                # "headless": False,
                 "args": [
                     "--disable-gpu",
                     "--disable-dev-shm-usage",
@@ -84,9 +84,8 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             
             # Add proxy settings if a proxy is specified
             if self.proxy:
-                proxy_settings = ProxySettings(server=self.proxy)
+                proxy_settings = {"server": self.proxy}
                 browser_args["proxy"] = proxy_settings
-                
                 
             self.browser = await self.playwright.chromium.launch(**browser_args)
             await self.execute_hook('on_browser_created', self.browser)
--- a/crawl4ai/async_webcrawler.py
+++ b/crawl4ai/async_webcrawler.py
@@ -24,9 +24,12 @@ class AsyncWebCrawler:
         crawler_strategy: Optional[AsyncCrawlerStrategy] = None,
         always_by_pass_cache: bool = False,
         verbose: bool = False,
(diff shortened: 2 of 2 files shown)


kylesf:
Are we going to take this in?

unclecode:
@neelthepatel8 Thx for the suggestion, however, `AsyncWebCrawler` already supports the proxy parameter. Or perhaps I am missing something, right now in the constructor of `AsyncWebCrawler` we set the proxy in line 51 `self.proxy = kwargs.get("proxy")`, and within the `start` function this code snippet set up the proxy:

```python
if self.proxy:
      proxy_settings = ProxySettings(server=self.proxy)
      browser_args["proxy"] = proxy_settings
```

Am I missing anything? 

@Barrierml already we have the proxy support, just simply pass the proxy when you are creating an instance from `AsyncWebCrawler` class.

neelthepatel8:
@unclecode My bad on re-setting the proxy in the `AsyncWebCrawlerStrategy` class, must have missed it. What I noticed though is that even though the `AsyncWebCrawler` can take in a proxy but it never uses it. I think the line 51 you mentioned is in `AsyncWebCrawlerStrategy`. So is it the case that you have to pass in the proxy only through the strategy?

unclecode:
@neelthepatel8 Not a problem at all, let's review and please help to make sure this works. Look at here `async_webcrawler.py`:

image


In this image line 28 I am passing `kwarg` to `AsyncPlaywrightCrawlerStrategy`, so technically, in the line 51 of `async_crawler_strategy.py` the passed `proxy` should be in the right place. Please cross check my answer, thx

neelthepatel8:
That makes sense, I didn't notice that my bad! 

We can close this PR then.

kylesf:
Yes, I miss understood another error to be a proxy error. My working proxy just times out when using it so I presumed it was not being used correctly. 

```
[LOG] 🌤️  Warming up the AsyncWebCrawler
[LOG] 🌞 AsyncWebCrawler is ready to crawl
[LOG] 🕸️ Crawling https://www.trychroma.com/engineering/serverless using AsyncPlaywrightCrawlerStrategy...
[ERROR] 🚫 arun(): Failed to crawl https://www.trychroma.com/engineering/serverless, error: [ERROR] 🚫 crawl(): Failed to crawl https://www.trychroma.com/engineering/serverless: Page.goto: Timeout 60000ms exceeded.
Call log:
navigating to "https://www.trychroma.com/engineering/serverless", waiting until "domcontentloaded"

2024-10-23 08:35:44 - app.helpers.web - INFO - Successfully extracted content
```

No issues without the proxy and its the same proxy used for multiple other things. Will just keep a pin in it and explore it later if people are successfully using the proxy feature. Thanks!

unclecode:
@kylesf @neelthepatel8 Just for the reference this is the way U can use the proxy:

```python
async def proxy():
    async with AsyncWebCrawler(headless = True, proxy_config = {
            "server": "http://ADDRESS",
            "username": "USR",
            "password": "PWD"
        }) as crawler:
        url = "https://www.nbcnews.com/business"
        result = await crawler.arun(
            url=url,
            bypass_cache=True,
            word_count_threshold = 10,
        )

    print("Done")
```

This is assuming you need username and password, if not do not pass it.

PR #128, opened 2024-10-03
spelling change in prompt

Darshan2104:
I was going through prompt and i encounter spelling mistake in prompt so helping with it :)


Natgho:
Attention here, please.

PR #176, opened 2024-10-17
fix requests version

mentaLwz:
change the `requests>=2.26.0,<2.32.3` to `requests>=2.26.0,<=2.32.3`  in requirements.txt

the previous version of requirements.txt is `requests==2.32.3`, so `<=2.32.3` is more precise?

https://github.com/unclecode/crawl4ai/commit/bccadec887214f33991ddebca9db2c18555c200a#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552L9

image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ numpy>=1.26.0,<2.1.1
 pillow==10.4.0
 playwright==1.47.0
 python-dotenv==1.0.1
-requests>=2.26.0,<2.32.3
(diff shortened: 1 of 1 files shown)


mentaLwz:
i think this solve the problem : https://github.com/unclecode/crawl4ai/pull/215#issue-2623372620

unclecode:
Thanks for the contribution. You're right, it actually fixed the problem; I appreciate that.

Issue #102, opened 2024-09-25
Crawl4ai Hard dependencites


TomerLankri:
Hi guys,
I started embedding your lib to my project, when i added it to requirements i started getting non stop errors on dependencies:
The conflict is caused by:
    The user requested fastapi==0.111.1
    crawl4ai 0.2.77 depends on fastapi==0.111.0

The conflict is caused by:
    The user requested pydantic==2.2.1
    fastapi 0.111.0 depends on pydantic!=1.8, !=1.8.1, !=2.0.0, !=2.0.1, !=2.1.0, <3.0.0 and >=1.7.4
    openai 1.11.0 depends on pydantic<3 and >=1.9.0
    pydantic-settings 2.0.3 depends on pydantic>=2.0.1
    crawl4ai 0.2.77 depends on pydantic==2.7.4

and its endless.
It seems like you added hard restrictions ("==") on libraries instead of ~= and its making it impossible to add to a project.

How can i resolve this?

example requierments.txt
```
alembic==1.11.3
apideck==1.2.3
apscheduler==3.10.4
cached-property==1.5.2
celery==5.3.5
cruft==2.15.0
email-validator==2.0.0.post2
emails==0.6
fastapi==0.111.0
greenlet==3.0.3
gunicorn==21.2.0
importlib-metadata==4.2.0
importlib-resources==5.12.0
mysql-connector-python==8.0.26
openai==1.11.0
passlib==1.7.4
plaid-python==15.5.0
postmarker==1.0
py-dotenv==0.1
pydantic==2.2.1
pydantic_core==2.6.1
pydantic-settings==2.0.3
pylint==2.17.5
pymysql==1.1.0
pytest-asyncio==0.23.5.post1
python-jose==3.3.0
python-multipart==0.0.7
pywatchman==1.4.1
sentry-sdk==1.43.0
stripe==6.0.0
uvicorn==0.22.0
numpy~=1.26.4
pandas~=2.2.2
requests~=2.31.0
scikit-learn~=1.5.1
tldextract==5.1.2
crawl4ai @ git+https://github.com/unclecode/crawl4ai.git
```

unclecode:
[...]

PR #215, opened 2024-10-30
build: make requirements more flexible

mjvankampen:
According to #102 the requirements specified are minimum version. Currently they are defined as fixed versions in requirements.txt and setup.py leading to projects consuming this package are limited to using exactly these requirements instead of a more flexible range. This PR addresses this.


unclecode:
@mjvankampen Thank you for your contribution to the library. Your pull request has been merged with love. Thanks again.

PR #229, opened 2024-11-04
Preventing NoneType has no attribute get Errors

bizrockman:
Sometimes the list contains Tag elements that do not have attrs set, resulting in this Error.

Change names of the docs to names that will work in Windows


unclecode:
Thank you so much for your collaboration. Your merge request is already accepted. I appreciate your support.

Issue #231, opened 2024-11-05
Incorrect Conversion of Relative to Absolute Paths for href in Web Pages


yulin0629:
The crawl4ai tool is not utilizing the provided base URL when converting relative paths to absolute paths for href attributes. Instead, it appears to be using some other method (possibly the domain or current page URL) for path resolution. This results in incorrect absolute URLs, potentially leading to broken links or inaccurate navigation within the crawled content.

unclecode:
@yulin0629 I would appreciate an example as I expect this to improve over time. If you can provide one example of your input, what you're getting, and what you expect, it would be very helpful. Thank you.

mindaugasrukas:
This example returns wrong URLs:
```
import crawl4ai

async def main():
    async with AsyncWebCrawler() as crawler:
        result = await crawler.arun(
            url="https://www.metabase.com/docs/latest/embedding/interactive-embedding",
        )
        print(result.markdown)

if __name__ == "__main__":
    asyncio.run(main())
```

The result looks similar to:
```
[ ](https://www.metabase.com/docs/latest/embedding/</>)
Product 
###### [Self-service analytics Business intelligence for everyone ](https://www.metabase.com/docs/latest/embedding/</product/business-analytics>) [ ![Embed](https://www.metabase.com/images/navigation-header/embed.svg) Embedded analytics Fast, flexible customer-facing analytics ](https://www.metabase.com/docs/latest/embedding/</product/embedded-analytics>)
Metabase Plans [ Starter and Open Source  ](https://www.metabase.com/docs/latest/embedding/</product/starter>) [ Pro  ](https://www.metabase.com/docs/latest/embedding/</product/pro>) [ Enterprise  ](https://www.metabase.com/docs/latest/embedding/</product/enterprise>)
Platform [ Data Sources  ](https://www.metabase.com/docs/latest/embedding/</data_sources/>) [ Security  ](https://www.metabase.com/docs/latest/embedding/</security/>) [ Cloud  ](https://www.metabase.com/docs/latest/embedding/</cloud/>)
[ Watch a 5-minute demo to see how to build a dashboard  ](https://www.metabase.com/docs/latest/embedding/</demo/>)
Features 
...
```

I narrowed down, that `urlparse.urljoin` receives wrong inputs:
```
base:  https://www.metabase.com/docs/latest/embedding/interactive-embedding
url: </product/business-analytics>
```

and `link_url` receives wrong a.href values:
```
[...]

PR #232, opened 2024-11-05
`fix: improve relative URL handling`

yulin0629:
Replaces manual URL path resolution with Python's urljoin for more robust and standardized URL normalization. This prevents edge cases and ensures compliance with RFC 3986 URL specifications.

The change simplifies maintenance while handling all relative path cases correctly, including fragments and parent directory traversal. 

Fixes: #231

### DIFF
--- a/crawl4ai/utils.py
+++ b/crawl4ai/utils.py
@@ -995,25 +995,17 @@ def normalize_url(href, base_url):
     if any(href.lower().startswith(proto) for proto in special_protocols):
         return href.strip()
         
-    # Handle anchor links
-    if href.startswith('#'):
-        return f"{base_url}{href}"
-        
-    # Handle protocol-relative URLs
-    if href.startswith('//'):
-        return f"{protocol}{href}"
-        
-    # Handle root-relative URLs
-    if href.startswith('/'):
-        return f"{protocol}//{domain}{href}"
-        
-    # Handle relative URLs
-    if not href.startswith(('http://', 'https://')):
-        # Remove leading './' if present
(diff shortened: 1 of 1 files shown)


yulin0629:
Thank you for sharing your `get_full_url` implementation! I've done some thorough testing comparing it with Python's built-in `urljoin`, and I'd like to share my findings.

While your implementation works for many cases, I found some edge cases where it behaves differently from the standard `urljoin`, particularly with relative paths when the base URL ends with a slash. Here's what I discovered:

```python
base_url = "https://example.com/news/"

# Case 1: Relative path
href = "relative/path.html?page=2"
urljoin:      "https://example.com/news/relative/path.html?page=2"
get_full_url: "https://example.com/relative/path.html?page=2"

# Case 2: Current directory path
href = "./current/path.html?page=2"
urljoin:      "https://example.com/news/current/path.html?page=2"
get_full_url: "https://example.com/current/path.html?page=2"
```

The main difference is that `urljoin` maintains the directory structure more accurately according to URL specifications. While your implementation works for your specific use cases, I would recommend using Python's built-in `urljoin` for the following reasons:

1. It correctly handles all edge cases with relative paths
2. It's part of the standard library and well-tested
3. It follows URL specifications more strictly
4. It's more maintainable as it's a standard solution

unclecode:
@yulin0629 Your pull request is already merged and available in the current version. For the next version, consider using a strategy design pattern to allow end-users to pass their own normalized function, providing more flexibility. I believe it's challenging to create a universal solution, but users can easily create their own strategies, especially when considering specific details.

PR #234, opened 2024-11-06
Fix typo: scrapper → scraper

devatnull:
Corrected "scrapper" to "scraper" in README to ensure it’s all about data extraction, not underground fights. Now crawl4ai can safely scrape data and not opponents..


unclecode:
Thank you. To be honest, it's been on my list for a while and I keep forgetting. So I just merged with pleasure :)).

PR #271, opened 2024-11-17
fix: crawler strategy exception handling and fixes

NanmiCoder:
Bug Description：

The following code captures exceptions but is missing the variable e. If this exception occurs, the program will lack the variable e and will panic.

https://github.com/unclecode/crawl4ai/blob/38044d4afe16b48783457e5ec8538e0a7592e16b/crawl4ai/crawler_strategy.py#L286-L289


ntohidikplay (approved):
yes, that's correct.

PR #294, opened 2024-11-26
Scraper uc

aravindkarnam:


aravindkarnam:
We are doing separation of concerns between the scraper and scheduler. In order to that there's going be radical changes in how the current scraper is implemented. Therefore closing this PR.

PR #314, opened 2024-12-04
fix: The extract method logs output only when self.verbose is set to True

lu4nx:


unclecode:
@lu4nx Thanks for the contribution. Your full request has been accepted and has been merged.

PR #332, opened 2024-12-09
 feat: Add remove_invisible_texts method to AsyncPlaywrightCrawlerStr…

nelzomal:
# New Feature
- Added functionality to detect and remove small or invisible text nodes.
- Logic is gated behind a remove_invisible_texts flag for optional use.

# High-Level Design 
Currently, the layout-related logic is implemented within the async_crawler_strategy, as layout details are best retrieved during the crawling phase when the web driver renders the page. This allows for efficient detection since the page is already fully rendered.

However, I propose saving the layout information during the crawling phase. This approach would enable us to leverage this data to implement more advanced heuristics during the subsequent scrape phase, providing greater flexibility and scalability.

I’m open to discussing this approach further before proceeding with the extended implementation.

# Implementation Details:
1. Utilizes document.createTreeWalker to traverse the DOM and process text nodes.
2. Checks parent elements of text nodes for invisibility based on:
- CSS properties (display, visibility, opacity).
- Element dimensions (width, height < 1px).
3. Includes specific handling for special cases: text within links (<a> tags) and tooltips (role="tooltip"):
- Only removed if parent element’s dimensions are too small. (website create very small link for accessibility purposes)

# Test
I couldn't run async test locally, so I create a small test script, and the result is
image

script:
```
import asyncio
import base64

from crawl4ai.cache_context import CacheMode


async def main():
    async with AsyncWebCrawler(verbose=True, headless=True) as crawler:
        urls = [
            
            "https://www.python.org",
            "https://www.stackoverflow.com",
            "https://developer.apple.com/documentation/photokit",
            "https://www.example.com"
        ]
       
        for j in range(len(urls)):
            print(f"staring url: {urls[j]}")
            for i in range(2):
                url = urls[j]
                wait_for = "css:.topictitle" if j == 2 else None
                remove_invisible_texts = (i == 0)
                print(f"remove_invisible_texts: {remove_invisible_texts}")
                result = await crawler.arun(
                    url=url,
                    remove_invisible_texts=remove_invisible_texts,
                    wait_for=wait_for,
                    cache_mode=CacheMode.BYPASS,
                )
            
                with open(f"result_{j}_{i}.md", "w") as f:
                    f.write(result.markdown)
            print("\n\n")
    


if __name__ == "__main__":
    asyncio.run(main())
```

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -1128,7 +1128,10 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             # Check for remove_overlay_elements parameter
             if kwargs.get("remove_overlay_elements", False):
                 await self.remove_overlay_elements(page)
-            
+           
+            if kwargs.get("remove_invisible_texts", False):
+                await self.remove_invisible_texts(page)
+
             html = await page.content()
             await self.execute_hook('before_return_html', page, html, context = context)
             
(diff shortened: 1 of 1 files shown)


aravindkarnam:
@nelzomal Sorry about the delay on this. Having contributed to the `filtering for small, invisible, or irrelevant images` feature, I absolutely agree with this idea. 

I can see that several changes have happened to `async_crawler_strategy.py` since you raised PR, but it hasn't impacted your logic. So let's proceed further. 

Could you resolve the merge conflicts, so I can proceed with reviewing your PR?

unclecode:
Hi, thanks for your contribution\! We're closing this PR as the codebase has changed significantly since it was opened and it now has merge conflicts with the current develop branch. If you'd like to re-submit against current develop, we'd be happy to review a fresh PR. Thanks for your effort\!

PR #337, opened 2024-12-10
Update async_webcrawler.py

lvzhengri:
add @asynccontextmanager 。

This is an innovative project, and I am very happy to make my contribution as a Chinese.


unclecode:
@lvzhengri Thanks for your kind words, and appreciation for the collaboration. So I will keep your comment in the Mandarin language as it is, as a memory of this collaboration from China. Thank you so much.

PR #357, opened 2024-12-19
Postpone legacy warning until logger is initialized

dvschuyl:
**TLDR**: potential circular code dependency in `AsyncWebCrawler.__init__`

#### Context:
When the newly added `browser_config` parameter is set in the `AsyncWebCrawler` initialization, it checks if other legacy parameters would also be set. In doing so, it logs a nice message. However, it uses a uninitialized `logger`. This logger depends on the `verbose` field of the `browser_config`.

#### Fix:
Postpone the warning to after the `logger` setup.

### DIFF
--- a/crawl4ai/async_webcrawler.py
+++ b/crawl4ai/async_webcrawler.py
@@ -76,12 +76,9 @@ class AsyncWebCrawler:
         """  
         # Handle browser configuration
         browser_config = config
+        legacy_warning = False
         if browser_config is not None:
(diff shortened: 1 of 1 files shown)


unclecode:
@dvschuyl Thank you for your contribution. In the upcoming version, I have removed this legacy message because it isn't neat and isn't needed anymore. You're right about the potential issue that may arise, so it's gone. That's why I won't merge it right now. In the new version, it won't be there anymore. Still, thank you so much, and please continue your contributions and support.

PR #419, opened 2025-01-06
chore: add py.typed file

Goldziher:
This PR adds the missing `py.typed` file. Without this file the library is considered untyped by type checkers.

### DIFF
--- a/crawl4ai/py.typed
+++ b/crawl4ai/py.typed
@@ -1,0 +1,1 @@
+# PEP-0561 typing sentinel, see: https://peps.python.org/pep-0561/
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
     license="MIT",
     packages=find_packages(),
     package_data={
-        'crawl4ai': ['js_snippet/*.js']
+        'crawl4ai': ['js_snippet/*.js', 'py.typed']
     },
     classifiers=[
         "Development Status :: 3 - Alpha",


midhunprahash-ui (changes requested):
from setuptools import setup, find_packages

setup(
    name="your_package_name",  # Replace with your package's name
    version="0.1",  # Adjust the version number
    license="MIT",
    packages=find_packages(),
    package_data={
        'crawl4ai': [
            'js_snippet/*.js',
            'py.typed'  # Make sure py.typed is included
        ],
    },
    classifiers=[
        "Development Status :: 3 - Alpha",
        # Add other classifiers as necessary
    ],
)

midhunprahash-ui (changes requested):
To correctly add a py.typed file to your package, you need to ensure the py.typed file is included in the package_data in your setup.py. However, the syntax in your snippet has a small issue. The dictionary keys in the package_data section should be unique, and you accidentally included the same key 'crawl4ai' twice.

Goldziher:
> To correctly add a py.typed file to your package, you need to ensure the py.typed file is included in the package_data in your setup.py. However, the syntax in your snippet has a small issue. The dictionary keys in the package_data section should be unique, and you accidentally included the same key 'crawl4ai' twice.
> 
> 

I think you are mistaken. The key is there only once. What you are seeing is a git diff showing before and after.

PR #458, opened 2025-01-15
Update getting-started.md

sennhvi:
'_v2' is what current version used.


aravindkarnam:
@sennhvi markdown_v2 is already marked for deprecation in v0.5 and it will be merged with markedown. Actually in this case documentation was ahead of code. So we can leave that docs untouched, once the release happens next week, this section will be accurate.

PR #462, opened 2025-01-15
fix: Add newline before pre codeblock start

jtanningbed:
Sometimes there is additional text before the 'pre' code block begins : 
image

This results in a malformed generated markdown: 
image

This just adds a newline before the start of the code block, should be fairly inconsequential. 

New result: 
image

There's another issue with handling whitespaces that are defined like: `<span class="w"> </span>`, seems like the parser is sending the entire content to `handle_data` if the code block is something like 
``` 
<code>
    "python"
    <span class="w"> </span>
    "main.py"
</code>
```
the html will be parsed with `handle_data("pythonmain.py")`, so the whitespace is not preserved. That seems like it would require updating the underlying HTMLParser logic or overriding it so I didn't try to address it myself. Just wanted to raise it in case it wasn't known. 
example: 
image

### DIFF
--- a/crawl4ai/html2text/__init__.py
+++ b/crawl4ai/html2text/__init__.py
@@ -1083,7 +1083,7 @@ class CustomHTML2Text(HTML2Text):
         # Handle pre tags
         if tag == 'pre':
             if start:
-                self.o('```\n')  # Markdown code block start
+                self.o('\n```\n')  # Markdown code block start
                 self.inside_pre = True
             else:
                 self.o('\n```\n')  # Markdown code block end


dmurat:
I would like to propose a more robust solution:

```
if tag == 'pre':
    if start:
        # Always start a new code block with a new line. Otherwise, it may not be rendered correctly.
        if not self.lastWasNL:
            self.o('\n')
        self.o('```\n')  # Markdown code block start
        self.inside_pre = True
    else:
        # Avoid adding unnecessary new lines at the end of the code block.
        if not self.lastWasNL:
            self.o('\n')
        self.o('```\n')  # Markdown code block end
        self.inside_pre = False
```

unclecode:
@jtanningbed @dmurat Thanks for sharing. This seems important to look. I will check it very soon.

unclecode:
Thank you for this fix! The missing newline before the opening code fence has been merged into develop. This is a small change but important for correct markdown rendering when text precedes a code block. Appreciate the contribution, even after all this time!

PR #581, opened 2025-01-29
Fix relative URL joining #570

### LINKED ISSUE #570: [Bug]: Relative Urls in the webpage not extracted properly
### crawl4ai version

0.4.247

### Expected Behavior

When parsing the markdown of a given webpage, 
1. if the href in the anchor/img/link tag is a relative url, it should be combined with base url properly (or let it remain relative) 
2. if the href in the anchor/img/link tag is a absolute url, it should not be combined with the base url:

The code to extract the Markdown:
```
    result = await crawler.arun(
        url=url,
    )
    return result.markdown
```

This is the expected Markdown of the webpage "https://docs.crawl4ai.com/"

```
[Crawl4AI Documentation (v0.4.3b2)](https://docs.crawl4ai.com/)
  * [ Home ](https://docs.crawl4ai.com/)
  * [ Quick Start ](https://docs.crawl4ai.com/core/quickstart/)
  * [ Search ](https://docs.crawl4ai.com/#)


  * Home
  * Setup & Installation
    * [Installation](https://docs.crawl4ai.com/core/installation/)
    * [Docker Deployment](https://docs.crawl4ai.com/core/docker-deploymeny/)
  * [Quick Start](https://docs.crawl4ai.com/core/quickstart/)
```

### Current Behavior

When parsing the markdown of a given webpage, relative urls are not being converted properly. relative urls are combined with the base url as base_url/<relative_url>  with the  angle brackets '<' and '>' symbols. 
Additionally, the relative url is being combined with base url even if the href contains absolute url.

This is the current Markdown of the webpage "https://docs.crawl4ai.com/"

```
[Crawl4AI Documentation (v0.4.3b2)](https://docs.crawl4ai.com/<https:/docs.crawl4ai.com/>)
  * [ Home ](https://docs.crawl4ai.com/<.>)
  * [ Quick Start ](https://docs.crawl4ai.com/<core/quickstart/>)
  * [ Search ](https://docs.crawl4ai.com/<#>)


  * Home
  * Setup & Installation
    * [Installation](https://docs.crawl4ai.com/<core/installation/>)
    * [Docker Deployment](https://docs.crawl4ai.com/<core/docker-deploymeny/>)
  * [Quick Start](https://docs.crawl4ai.com/<core/quickstart/>)
 ```

Side note: spelling mistake in ```https://docs.crawl4ai.com/core/docker-deploymeny/```

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
URL: https://docs.crawl4ai.com/
```

### Steps to Reproduce

```bash
1. Run the below code snippet for the mentioned url
   python webpage_crawler.py https://docs.crawl4ai.com/ crawl4ai.md
2. Compare the generated Markdown and the raw html.
```

### Code snippets

```python
# filename: webpage_crawler.py

import asyncio

async def get_markdown(url: str) -> str:
        try:
            result = await crawler.arun(
                url=url,
            )
            if result.url != url:
                print(f"Redirected to {result.url}")
            if not result.success:
                raise Exception(result.error_message)
            if result.status_code == 404:
                raise Exception(f"url not found")
            return result.markdown
        except Exception as err:
            print("Crawler failed for", url)
            raise err
        

async def get_cleaned_html(url: str) -> str:
        try:
            result = await crawler.arun(
                url=url,
            )
            if result.url != url:
                print(f"Redirected to {result.url}")
            if not result.success:
                raise Exception(result.error_message)
            if result.status_code == 404:
                raise Exception(f"url not found")
            return result.cleaned_html
        except Exception as err:
            print("Crawler failed for", url)
            raise err

if __name__ == "__main__":
    import sys
    
    if len(sys.argv) != 3:
        print("Usage: python script.py <url> <output_file>")
        sys.exit(1)
    
    url = sys.argv[1]
    output_file = sys.argv[2]
    
    markdown = asyncio.run(get_markdown(url))
    print(len(markdown))
    with open(output_file, 'w', encoding='utf-8') as f:
        f.write(markdown)
```

### OS

Windows 10 (also observed on Linux)

### Python version

3.11

### Browser

Google Chrome

### Browser version

131.0.6778.265

_No response_

Sparshsing:
## Summary
`Fixes #570` 
Relative url joining has been fixed. The issue was due to incorrect usage of link protection using <> tags.

crawl4ai/html2text.py - 
While parsing the "a" tags in handle_tag method, the url in "href" attr was being protected by surrounding with <> in the opening tag, and then in the closing tag it was joined with base_url. To fix this, the protetion with <> is done after the url is joined with the base_url.
The same process has been done in the o() method.

crawl4ai\markdown_generation_strategy.py -  
In the conversion of links to citations for markdown_v2, we remove the <> symbols from the url if it was protected, then we join the url with base_url. 
Also the fast_urljoin method was modified, since it would join a url incorrectly if the base url included a path eg.  https://docs.crawl4ai.com/core/crawler-result/.

## How Has This Been Tested?
It was tested by generating markdown and markdown_v2 of few urls. 
eg. https://docs.crawl4ai.com/, https://docs.crawl4ai.com/core/crawler-result/
The links in the fixed markups were working as expected.

## Checklist:

### DIFF
--- a/crawl4ai/html2text/__init__.py
+++ b/crawl4ai/html2text/__init__.py
@@ -503,8 +503,7 @@ class HTML2Text(html.parser.HTMLParser):
                 self.o(self.close_quote)
             self.quote = not self.quote
 
-        def link_url(self: HTML2Text, link: str, title: str = "") -> None:
-            url = urlparse.urljoin(self.baseurl, link)
+        def format_markdown_link(self: HTML2Text, url: str, title: str = "") -> None:
             title = ' "{}"'.format(title) if title.strip() else ""
             self.o("]({url}{title})".format(url=escape_md(url), title=title))
 
@@ -521,8 +520,6 @@ class HTML2Text(html.parser.HTMLParser):
                     self.astack.append(attrs)
                     self.maybe_automatic_link = attrs["href"]
                     self.empty_link = True
-                    if self.protect_links:
-                        attrs["href"] = "<" + attrs["href"] + ">"
                 else:
                     self.astack.append(None)
             else:
@@ -540,7 +537,10 @@ class HTML2Text(html.parser.HTMLParser):
                             self.p_p = 0
                             title = a.get("title") or ""
                             title = escape_md(title)
-                            link_url(self, a["href"], title)
+                            joined_url = urlparse.urljoin(self.baseurl, a["href"])
+                            if self.protect_links:
+                                joined_url = "<" + joined_url + ">"
+                            format_markdown_link(self, joined_url, title)
                         else:
                             i = self.previousIndex(a)
                             if i is not None:
@@ -838,11 +838,14 @@ class HTML2Text(html.parser.HTMLParser):
                 newa = []
                 for link in self.a:
                     if self.outcount > link.outcount:
+                        joined_url = urlparse.urljoin(self.baseurl, link.attrs["href"])
+                        if self.protect_links:
+                            joined_url = "<" + joined_url + ">"
                         self.out(
                             "   ["
                             + str(link.count)
                             + "]: "
-                            + urlparse.urljoin(self.baseurl, link.attrs["href"])
+                            + joined_url
                         )
                         if "title" in link.attrs and link.attrs["title"] is not None:
                             self.out(" (" + link.attrs["title"] + ")")
--- a/crawl4ai/markdown_generation_strategy.py
+++ b/crawl4ai/markdown_generation_strategy.py
@@ -14,11 +14,6 @@ def fast_urljoin(base: str, url: str) -> str:
     """Fast URL joining for common cases."""
     if url.startswith(("http://", "https://", "mailto:", "//")):
         return url
-    if url.startswith("/"):
-        # Handle absolute paths
-        if base.endswith("/"):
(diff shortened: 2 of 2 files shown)


Sparshsing:
Maybe there should be a check at the top level code to add / at the end of url if not exist, as it can cause issues with relative urls.

aravindkarnam:
@Sparshsing While your diagnosis of this issue is correct, we would like to handle this by [simply making "protect_links" false by default](https://github.com/unclecode/crawl4ai/commit/f982875040316800d8be296b436c47578643af9a). The original intention behind this is to protect URLs that contain special characters or segments that could be misinterpreted by Markdown processors. For example some Markdown processors might interpret the underscore as emphasis markup. (so a link like https://docs.example.com/user_guide/installation/ can be rendered incorrectly)

However as this is obviously causing some confusions, we'll turn this off by default and users who care about markdown validity in grammar and syntax can turn this on.

Closing your PR, since it does some deeper changes to html2text package that we want to avoid

Sparshsing:
Thanks @aravindkarnam for viewing this. I agree the simpler approach is more suitable and less risky.

PR #640, opened 2025-02-08
Fix cdp_url import into the managed browser

mikamind:
## Summary

I've squashed a bug where the cdp_url was being ignored, even when it was clearly specified in the BrowserConfig.
Now, the managed browser gets the proper cdp_url injection it needs for any remote connections.

async_crawler_strategy.py - Updated the ManagedBrowser initialization to ensure the `cdp_url` from the BrowserManager's config is passed along. 

## How Has This Been Tested?

Simply pass any valid cdp_url into the BrowserConfig. Previously, it was being overlooked. With this fix, the URL is now correctly passed along, confirming our solution works as intended.

## Checklist:

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -349,6 +349,7 @@ class BrowserManager:
                 headless=self.config.headless,
                 logger=self.logger,
                 debugging_port=self.config.debugging_port,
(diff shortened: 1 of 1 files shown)


hakzarov:
why not accepted? It's literally a one line fix. I found this PR after searching for why my crawler doesn't use CDP...

PR #681, opened 2025-02-15
executing JS should happen after waiting

ksallee:
## Summary

If JS code execution happens before `delay_before_return_html`, if any, then some elements might not be loaded in the page.

crawl4ai/async_crawler_strategy.py to do exactly that.

## How Has This Been Tested?
Tested with delay_before_return_html and a page that requires a bit of loading, failed before change, worked after change.


unclecode:

PR #777, opened 2025-03-02
Update no-llm-strategies.md

PhySci:
## Summary
Amend example: rename `llm_provider` to `provider` because of https://github.com/unclecode/crawl4ai/blob/3b1025abbb6e2565602c05f9a959458da3531f3a/crawl4ai/extraction_strategy.py#L1068C9-L1068C17

docs/md_v2/extraction/no-llm-strategies.md

## How Has This Been Tested?
I have tested it locally. Now example works for other OpenAI models. It allows to process large documents.

## Checklist:

- [+ ] My code follows the style guidelines of this project
- [+ ] I have performed a self-review of my own code
- [ +] I have commented my code, particularly in hard-to-understand areas
- [+] I have made corresponding changes to the documentation
- [ +] I have added/updated unit tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/md_v2/extraction/no-llm-strategies.md
+++ b/docs/md_v2/extraction/no-llm-strategies.md
@@ -434,7 +434,7 @@ html = """
 css_schema = JsonCssExtractionStrategy.generate_schema(
     html,
     schema_type="css",  # This is the default
-    llm_provider="openai/gpt-4o",  # Default provider
+    provider="openai/gpt-4o",  # Default provider
     api_token="your-openai-token"  # Required for OpenAI
 )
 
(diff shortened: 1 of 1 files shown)


aravindkarnam:
@PhySci Actually in the upcoming version v0.5, we are introducing a new config for setting up connections to LLM Providers, called `LlmConfig`, which provides user with utilities to easily load auth keys easily from the env variables etc and also to reuse it across different, LLM related functionalities like extraction, filtering and schema generation strategies. 

So this particular commit would not be inline with the changes in upcoming release. Hence closing the PR.

Issue #778, opened 2025-03-03
[Bug]: Failed to get ssl certificate because it was not serialized correctly


wakaka6:
### crawl4ai version

0.4.3b3

### Expected Behavior

API return ssl certification

### Current Behavior

client:
Internal Server Error

server:
"POST /crawl_direct HTTP/1.1" 500 Internal Server Error

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
curl -X 'POST' \
  'http://192.168.1.2:11235/crawl_direct' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "urls": "https://example.com",
  "magic": true,
  "extra": {
      "page_timeout": 30000,
      "proxy": "",
      "scan_full_page": true,
      "fetch_ssl_certificate": true,
      "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:92.0) Gecko/20100101 Firefox/92.0"
    },
  "cache_mode": "disabled"
}'
```

### Steps to Reproduce

```bash

```

### Code snippets

```python

```

### OS

Linux

### Python version

3.12

### Error logs & Screenshots (if applicable)

```
INFO:     192.168.1.2:56052 - "POST /crawl_direct HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/iboomw/mambaforge/envs/crawl4ai/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iboomw/mambaforge/envs/crawl4ai/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iboomw/mambaforge/envs/crawl4ai/lib/python3.12/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/home/iboomw/mambaforge/envs/crawl4ai/lib/python3.12/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
[...]

PR #864, opened 2025-03-21
feat(ssl-certificate): get ssl certificate support proxy

wakaka6:
## Summary

Support proxy when getting ssl certificate

```py
import asyncio
from crawl4ai import (
    AsyncWebCrawler,
    BrowserConfig,
    CrawlerRunConfig,
    CacheMode,
    DefaultMarkdownGenerator,
    CrawlResult,
)
from crawl4ai.configs import ProxyConfig


async def main():
    browser_config = BrowserConfig(headless=True, verbose=True)
    async with AsyncWebCrawler(config=browser_config) as crawler:
        crawler_config = CrawlerRunConfig(
            cache_mode=CacheMode.BYPASS,
            magic=True,
            fetch_ssl_certificate=True,
            proxy_config=ProxyConfig(server="socks5://127.0.0.1:1088"),
            markdown_generator=DefaultMarkdownGenerator(
                # content_filter=PruningContentFilter(
                #     threshold=0.48, threshold_type="fixed", min_word_threshold=0
                # )
            ),
        )
        result : CrawlResult = await crawler.arun(
            url="https://www.google.com", config=crawler_config
        )
        print("ssl:", result.ssl_certificate)
        print("markdown: ",result.markdown[:500])


if __name__ == "__main__":
    asyncio.run(main())

```

image

## List of files changed and why
ssl_ceritficate.py 
- Support proxy when getting ssl certificate
- Support export certificate to playwright format with `ssl_ceritificate.to_playwright_format()`
- Support `str(ssl_ceritificate)`

proxy_config.py
- Support for conversion of URLs with embedded credentials to ProxyConfig. The user and password in the URL with embedded credentials overrides `self.username` and `self.password`.
- e.g. 
  ```
  ProxyConfig(server="http://user:pass@proxy-server:1080",username="", password="")
  --(normalize)--> ProxyConfig(server="http://proxy-server:1080",username="user", password="pass")
  ```

async_crawler_strategy.py
- Crawling will set the proxy according to the configuration.

## How Has This Been Tested?

- In the environment of network limitation, use http, https and socks5 proxy to test the website which is banned by firewall(like GFW), all of them can get SSL certificate(e.g. you can't access google directly in China, you need external proxy).
- In the environment where there is no network restriction, you can also get the certificate without using proxy.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
[...]

PR #961, opened 2025-04-09
feat(ssl-certificate): get ssl certificate support proxy

wakaka6:
## Summary

Support proxy when getting ssl certificate

Same as #864 PR, but based on next branches.

```py
import asyncio
from crawl4ai import (
    AsyncWebCrawler,
    BrowserConfig,
    CrawlerRunConfig,
    CacheMode,
    DefaultMarkdownGenerator,
    CrawlResult,
)
from crawl4ai.configs import ProxyConfig


async def main():
    browser_config = BrowserConfig(headless=True, verbose=True)
    async with AsyncWebCrawler(config=browser_config) as crawler:
            cache_mode=CacheMode.BYPASS,
            magic=True,
            fetch_ssl_certificate=True,
            proxy_config=ProxyConfig(server="socks5://127.0.0.1:1088"),
            markdown_generator=DefaultMarkdownGenerator(
                # content_filter=PruningContentFilter(
                #     threshold=0.48, threshold_type="fixed", min_word_threshold=0
                # )
            ),
        )
        result : CrawlResult = await crawler.arun(
            url="https://www.google.com", config=crawler_config
        )
        print("ssl:", result.ssl_certificate)
        print("markdown: ",result.markdown[:500])


if __name__ == "__main__":
    asyncio.run(main())

```

image

ssl_ceritficate.py 
- Support proxy when getting ssl certificate
- Support export certificate to playwright format with `ssl_ceritificate.to_playwright_format()`
- Support `str(ssl_ceritificate)`

proxy_config.py
- Support for conversion of URLs with embedded credentials to ProxyConfig. The user and password in the URL with embedded credentials overrides `self.username` and `self.password`.
- e.g. 
  ```
  ProxyConfig(server="http://user:pass@proxy-server:1080",username="", password="")
  --(normalize)--> ProxyConfig(server="http://proxy-server:1080",username="user", password="pass")
  ```

async_crawler_strategy.py
- Crawling will set the proxy according to the configuration.

## How Has This Been Tested?

- In the environment of network limitation, use http, https and socks5 proxy to test the website which is banned by firewall(like GFW), all of them can get SSL certificate(e.g. you can't access google directly in China, you need external proxy).
- In the environment where there is no network restriction, you can also get the certificate without using proxy.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dependencies = [
     "faust-cchardet>=2.1.19",
     "aiohttp>=3.11.11",
     "humanize>=4.10.0",
+    "PySocks @ git+https://github.com/amirasaran/PySocks.git@3da955fd212ce02c3ab3bc166b5bfac3c91b4019"
 ]
 classifiers = [
     "Development Status :: 4 - Beta",
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,3 +22,4 @@ nltk>=3.9.1
 rich>=13.9.4
 cssselect>=1.2.0
 faust-cchardet>=2.1.19
+PySocks @ git+https://github.com/amirasaran/PySocks.git@3da955fd212ce02c3ab3bc166b5bfac3c91b4019
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -31,6 +31,7 @@ from types import MappingProxyType
 import contextlib
 from functools import partial
 
+
 class AsyncCrawlerStrategy(ABC):
     """
     Abstract base class for crawler strategies.
@@ -40,6 +41,7 @@ class AsyncCrawlerStrategy(ABC):
     @abstractmethod
     async def crawl(self, url: str, **kwargs) -> AsyncCrawlResponse:
         pass  # 4 + 3
+
 
 class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
     """
@@ -490,7 +492,7 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
         response_headers = {}
         execution_result = None
         status_code = None
-        redirected_url = url 
+        redirected_url = url
 
         # Reset downloaded files list for new crawl
         self._downloaded_files = []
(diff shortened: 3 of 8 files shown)


pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
     "faust-cchardet>=2.1.19",
     "aiohttp>=3.11.11",
     "humanize>=4.10.0",
+    "PySocks @ git+https://github.com/amirasaran/PySocks.git@3da955fd212ce02c3ab3bc166b5bfac3c91b4019"
```
aravindkarnam: @wakaka6 Could you explain why you're using a direct Git reference with a specific commit hash instead of the official PyPI package?

requirements.txt
```diff
@@ -21,4 +21,5 @@ psutil>=6.1.1
 nltk>=3.9.1
 rich>=13.9.4
 cssselect>=1.2.0
-faust-cchardet>=2.1.19
\ No newline at end of file
+faust-cchardet>=2.1.19
+PySocks @ git+https://github.com/amirasaran/PySocks.git@3da955fd212ce02c3ab3bc166b5bfac3c91b4019
```
aravindkarnam: Could you explain why you're using a direct Git reference with a specific commit hash instead of the official PyPI package?

aravindkarnam (changes requested):
@wakaka6 @unclecode  I'm not very familiar with SSL certificate related concepts or PySocks library. So I'm not in a position to properly review this right away, but I'll do by next alpha release. 

I just had a question w.r.t the dependancy, please check that out.

requirements.txt
wakaka6: The official Pysocks has buggy(https://github.com/Anorov/PySocks/pull/147) support for http proxies and the original author doesn't seem to have maintained it for a long time as well, so i need to use a specific version of Pysocks as a dependency.

aravindkarnam:
Fixes: #778 - Tagging for visibility

requirements.txt
wakaka6: I read the [mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Proxy-Authorization) carefully and it seems that the http basic authentication should ignore case, so it's a problem with my proxy program, so we can just use the official version of pysocks.

wakaka6:
@aravindkarnam When can we get it to merge? I think I'm ready.

ntohidi:
this is fixed in the latest version, 0.7.6

PR #891, opened 2025-03-26
fix: tests and associated bugs

stevenh:
This is a placeholder which contains all the fixes and improvements needed to get all the existing tests running error free.
    
There are a small number of tests which have been marked as skipped due to flaky behaviour or significant functionality needed to make them work.
    
The intention is to split this out into many more consumable changes.

This should not be merged as is.

PR #969, opened 2025-04-10
fix: caching, downloads, js_code and browser manager

stevenh:
## Summary

Fixes missing metadata from page cache, incomplete file downloads, the handling of js_code and playwright browser reuse, alongside adjacent issues in the affected files, see below for more details.

## Details

Fix the handling of file downloads in AsyncPlaywrightCrawlerStrategy which wasn't waiting for the download to complete before returning, which resulted in race conditions and incomplete or missing downloads.

Fix incorrect field use for headers and user_agent in AsyncPlaywrightCrawlerStrategy.

Fix the type used for the logger in AsyncPlaywrightCrawlerStrategy to the base class Logger. This allows the logger to be used correctly and prevents type errors when using the logger.

Add missing abstract methods to AsyncCrawlerStrategy and implement in its subclasses.

Fix the extraction of the error message when processing console error log in log_console.

Fix the handling of js_code in AsyncPlaywrightCrawlerStrategy to ensure that the code is executed correctly and the result is returned when the script starts with a variable or constant declaration.

Remove unused time variables and commented out code to improve readability.

Remove duplicate Exception handling.

Fix the handling of raw requests in AsyncHTTPCrawlerStrategy which was previously using the parsed path corrupting the request.

Fix the browser manager to ensure that it always returns a valid browser instance. This eliminates the class level playwright instance which wasn't being cleaned up on close, resulting in subsequent callers receiving a closed instance and causing errors.

Fix the storage of links, media and metadata to ensure that the correct values are stored and returned. This prevents incorrect results when using the cached results.

Use Field for default values in Media, Links and ScrapingResult pydantic models to prevent invalid results.

Handle undefined string passed to MediaItem width.

Add parameters to the CrawlResult.__init__ method to provide type hints for the fields, which provides better type checking.

Fix database query retry handling for the case where the database table is missing when using the SQLite database. This prevents the crawler from failing when the table is not present and allows it to continue crawling.

This prevents failing tests after one which drops the database table.

Remove commit call for queries which don't need it.

Sync table definition in legacy code so prevent mixed legacy and new code tests from conflicting with each other.

Fix the caching of markdown field in DB / files which was only storing the single value, which caused failures when using cached results.

Export the markdown field in StringCompatibleMarkdown, so we don't need to use a private field to ensure that the value is serialised correctly.

Correctly initialise BrowserConfig and CrawlerRunConfig from kwargs to ensure legacy parameters work as expected.

Add ruff settings to pyproject.toml to prevent deprecation warnings beyond our control and to disable formatting for now to avoid the noise it would generate.

## How Has This Been Tested?

This has been tested both separately with the updated tests as well as in combination with the wider tests as part of https://github.com/unclecode/crawl4ai/pull/891

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,3 +96,24 @@ crawl4ai = { workspace = true }
 dev = [
     "crawl4ai",
 ]
+
+[tool.pytest.ini_options]
+asyncio_default_fixture_loop_scope = "function" # Prevent deprecation warning
+# Disable deprecation warnings for code we don't control.
+filterwarnings = [
+    "ignore::DeprecationWarning",
+    "default::DeprecationWarning:__main__",
+    "default::DeprecationWarning:crawl4ai.*",
+]
+timeout = 20
+timeout_func_only = true
+
+# Basic configuration for the `ruff` tool to preserve basic formatting.
+[tool.ruff]
+line-length = 120
+target-version = "py39"
+
+# As a temporary workaround, exclude all files from ruff formatting as the
+# code base has various code style and would introduce a lot of noise.
+[tool.ruff.format]
+exclude = ["*.py"]


stevenh:
Closing as never got any traction, so we've moved away from crawl4ai.

If someone wants to pick up the branch and reuse, feel free.

Issue #884, opened 2025-03-25
[Bug]: maximum recursion depth exceeded while calling a Python object


yumingmin88:
### crawl4ai version

0.5.0.post4

### Expected Behavior

Sequential execution does not produce an exception

### Current Behavior

Hello, the following error occurred when crawling 3000 urls using the following code：maximum recursion depth exceeded while calling a Python object
The code used is changed from issue399:
[https://github.com/unclecode/crawl4ai/issues/399](url)
```
async def crawl_url_content(url):
    res = ''

    logger.info("\n=== Single URL Crawling with Browser Reuse + Memory Check ===")

    # We'll keep track of peak memory usage
    peak_memory = 0
    process = psutil.Process(os.getpid())

    def log_memory(prefix: str = ""):
        nonlocal peak_memory
        current_mem = process.memory_info().rss  # in bytes
        if current_mem > peak_memory:
            peak_memory = current_mem
        logger.info(f"{prefix} Current Memory: {current_mem // (1024 * 1024)} MB, Peak: {peak_memory // (1024 * 1024)} MB")

    ua = UserAgent()
    # Minimal browser config
    browser_config = BrowserConfig(
        headless=True,
        verbose=False,   
        extra_args=["--disable-gpu", "--disable-dev-shm-usage", "--no-sandbox"],
        user_agent_generator_config={"mode": "random"},
        java_script_enabled=True,
        user_agent=ua.random
    )
    crawl_config = CrawlerRunConfig(
        cache_mode=CacheMode.BYPASS, 
        page_timeout=30000,
        excluded_tags=EXCLUDE_TAGS,
        check_robots_txt=True,
    )

    # Create the crawler instance
    crawler = AsyncWebCrawler(config=browser_config)
    await crawler.start()

    try:
        success_count = 0
        fail_count = 0
        
        # Check memory usage prior to processing
        log_memory(prefix="Before processing: ")

        # Process single URL
        session_id = "single_url_session"
        result = await crawler.arun(url=url, config=crawl_config, session_id=session_id)

        # Check memory usage after processing
        log_memory(prefix="After processing: ")

        # Evaluate result
        if isinstance(result, Exception):
            logger.info(f"Error crawling {url}: {result}")
            fail_count += 1
        elif result.success:
            success_count += 1
        else:
            fail_count += 1
        res = result

        logger.info(f"\nSummary:")
        logger.info(f"  - Successfully crawled: {success_count}")
[...]

PR #977, opened 2025-04-12
Prevent pseudo-memory leak from repeated colorama stream wrapping

claudioarione:
## Summary
This commit fixes Bug #884 by introducing a safe initialization for the colorama library. Previously, the code called colorama’s *init()* method every time an instance of *AsyncWebCrawler*, *BrowserProfiler*, or *AsyncLogger* was created, as well as each time the utility function *create_box_message()* was invoked. In practice, creating multiple instances (which, although not best practice, can still occur) or printing many errors resulted in many redundant calls to *init()*.

This is problematic because each call to *init()* wraps the standard output/error streams (e.g., sys.stdout and sys.stderr) with an additional layer of colorama’s stream wrapper. With repeated calls, the streams become nested within each other, leading to a cascading chain of wrappers. This not only increases the call stack—potentially causing recursion errors—but also leads to a pseudo-memory leak as more and more wrapper objects are created and retained in memory.

This PR replaces the direct call to colorama’s *init()* with a custom *init_colorama_safe()* function. This function ensures that colorama’s *init()* is executed only once, thereby avoiding the excessive wrapping of streams and the associated resource consumption.

**Note:**  
The current implementation relies on a configuration variable, `USE_COLORED_SHELL`, which defaults to `True`. If set to `False`, colorama initialization is skipped, leaving stdout and stderr untouched. Although it would be straightforward to tie the default to an environment variable (e.g., `CRAWL4AI_USE_COLORED_SHELL`), I chose not to for two primary reasons: (1) no other configuration option in this repo is set up this way, implying that manual editing of this file or adjustment by the user is acceptable; and (2) I have received feedback from only one user requesting unwrapped streams, which suggests it might not be a priority for most.

## List of Files Changed and Why
- **utils.py:** Contains the newly implemented *init_colorama_safe()* function, which includes a guard condition to ensure *init()* is only called once; also, the function *create_box_message()* now calls *init_colorama_safe()*.
- **config.py:** Introduces the new configuration variable `USE_COLORED_SHELL`.
- **async_logger.py, browser/profiles.py, browser_profiler.py:** Now call *init_colorama_safe()* instead of directly calling colorama’s *init()*.

## How Has This Been Tested?
I ran the project on Linux, macOS, and Windows. The logs are produced as expected, and diagnostic stack traces (from both crawl4ai and colorama functions) confirm that the streams are not wrapped multiple times. See below for an explanation of the pre-existing tests.

## Checklist:
- [✅] My code follows the style guidelines of this project
- [✅] I have performed a self-review of my own code
- [✅] I have commented my code, particularly in hard-to-understand areas
- [❌] I did not make corresponding changes to the documentation — *no need*
- [🟡] I have added/updated unit tests that prove my fix is effective or that my feature works
- [🟡] New and existing unit tests pass locally with my changes

*Explanation of 🟡:*  
The tests folder contains many tests that refer to previous versions of the code. I have made some adjustments locally; however, if a comprehensive new test campaign is required, I might need to review and update the tests more thoroughly.


ftnext:
This pull request is very much appreciated.

colorama 0.4.6 (latest version) added `just_fix_windows_console()`.
https://github.com/tartley/colorama/tree/0.4.6?tab=readme-ov-file#initialisation
>It's safe to call this function multiple times. It's safe to call this function on non-Windows platforms, but it won't do anything.

I think `init_colorama_safe()` should call this instead of `init()`.

PR #1081, opened 2025-05-06
fix: corrected formula and precomputed values

### LINKED ISSUE #1080: [Bug]: Wrong scoring with depth scorer
### crawl4ai version

crawl4ai-0.6.2

### Expected Behavior

When using the Best First Crawling Strategy together with the path depth scorer, while setting the optimal_depth to 0, I expect the crawler to first crawl the URLs with the shortest path. This would make sense because they are closer to the optimal depth.

### Current Behavior

It is doing the exact opposite: it is first crawling the URLs with the longest depth.

`bff_strategy.py` states the following: "Lower scores are treated as higher priority."

The `PathDepthScorer` calculates the score like this: `1.0 / (1.0 + distance)`.
A distance of 0 from the optimal depth would therefore result in a score of 1, while a distance of 9 from the optimal depth would result in a score of 0.1 — which, by definition, would be considered a higher priority.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python
import asyncio
from crawl4ai import *
from crawl4ai.deep_crawling import BestFirstCrawlingStrategy
from crawl4ai.deep_crawling.scorers import PathDepthScorer

async def main():
    path_depth_scorer = PathDepthScorer(optimal_depth=0)

    config = CrawlerRunConfig(
            deep_crawl_strategy=BestFirstCrawlingStrategy(
                url_scorer=path_depth_scorer,
                max_depth=20
            ),
            stream=True,
        )

        async for result in await crawler.arun("https://docs.crawl4ai.com/", config=config):
            if result.success:
                print(f"✅ Scraped {result.url}")


if __name__ == "__main__":
    asyncio.run(main())
```

### OS

Windows

### Python version

3.12.0

### Browser

_No response_

### Browser version

_No response_

_No response_

Joorrit:
## Summary
Fixes #1080 

Corrected the scoring function of the `PathDepthScorer` as it was giving a path that has a higher distance to the optimal depth a lower score and a path that is close to the optimal depth a higher score.

Also updated the precomputed values that are only used by the `PathDepthScorer`.

## How Has This Been Tested?
I testet the code before and after this change and it works as expected after the change.

## Checklist:

### DIFF
--- a/crawl4ai/deep_crawling/scorers.py
+++ b/crawl4ai/deep_crawling/scorers.py
@@ -10,8 +10,8 @@ import ctypes
 import platform
 PLATFORM = platform.system()
 
-# Pre-computed scores for common year differences
-_SCORE_LOOKUP = [1.0, 0.5, 0.3333333333333333, 0.25]
+# Pre-computed scores for path distance differences
+_SCORE_LOOKUP = [0.0, 0.5, 0.6666666666666667, 0.75]
 
 # Pre-computed scores for common year differences
 _FRESHNESS_SCORES = [
@@ -241,8 +241,8 @@ class PathDepthScorer(URLScorer):
         
         if distance < 4:
             return _SCORE_LOOKUP[distance]
-            
-        return 1.0 / (1.0 + distance)                                             
+
+        return 1.0 - 1.0 / (1.0 + distance)                                            
 
 class ContentTypeScorer(URLScorer):
     __slots__ = ('_weight', '_exact_types', '_regex_types')


unclecode:
Thanks for working on this — the scorer inversion is a real issue. However, the PR currently only updates `bfs_strategy.py` and `scorers.py`, which would break the other two deep crawl strategies that still expect "higher is better" semantics.

**What needs to be fixed before we can merge:**

1. **`crawl4ai/deep_crawling/dfs_strategy.py`** (line ~308):
   - `score < self.score_threshold` → `score > self.score_threshold`
   - The default `score_threshold` needs to change from `-infinity` to `infinity`
   - The log message should say "above threshold" not "below threshold"
   - Line ~320: `valid_links.sort(key=lambda x: x[1], reverse=True)` → `reverse=False` (since lower is now better)

2. **`crawl4ai/deep_crawling/bff_strategy.py`** (line ~226, ~308):
   - Uses `-new_score` for priority queue ordering (negates to get highest first). With inverted scores (lower = better), this should just be `new_score` (no negation), since `PriorityQueue` already pops the lowest value first.

3. **Tests needed:**
   - Run the existing scorer tests: `pytest tests/general/test_deep_crawl_scorers.py -v`
   - Add a test that exercises BFS, DFS, and BFF with a `KeywordRelevanceScorer` + `score_threshold` to confirm all three strategies correctly skip/include URLs under the new semantics
   - Verify that the priority ordering in BFF actually puts the best-scoring URLs first

Once those are addressed, we'd be happy to merge this. The scorer logic fix itself is correct — it just needs to be applied consistently across all strategies.

unclecode:

PR #1094, opened 2025-05-10
feat: token bucket rate limiting

stevenh:
## Summary
Replace the existing rate limiting mechanism with a token bucket algorithm to improve request handling efficiency and control.
    
The previous implementation used a simple last request and current delay calculation, which could lead to uneven request distribution when multiple requests were made in quick succession. The result of this was the more links discovered by as single request the more likely it was that we would trigger 429 and 503 response codes, which combined with max_retries would cause the crawler to fail to successfully process all pages.

The new implementation uses a token bucket algorithm, which allows for more flexible and efficient rate limiting. The token bucket algorithm allows for bursts of requests while maintaining an average rate over time. This combined with active rate limit header processing results in a much higher success rate for crawls.

In addition the option to disable max retry limit, the new default, has been tested to confirm 100% success rate for crawls on a site which performs active rate limiting and previously resulted in a large number of failures.

Finally the rate limiter starts with default rate which will be adjusted per domain based on the rate limit headers returned by the server.

## How Has This Been Tested?
New set of tests added to tests/test_async_dispatcher.py

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code


stevenh:
This builds on top of other changes, which haven't yet been merged so would need that done first, hence leaving as draft.

stevenh:
Closing as never got any traction, so we've moved away from crawl4ai.

If someone wants to pick up the branch and reuse, feel free.

PR #1102, opened 2025-05-12
fix(crawler): use base_url when href is absolute

pawel-twardziak:
## Summary

this fixes an issue, when `url` is `http://domain.com/path` and a href on this page is `/path-2` resulting with `http://domain.com/path/path-2` which is wrong


pawel-twardziak:
@unclecode I see I changed the code you have created. Please, look at it.

PR #1157, opened 2025-05-26
feat(crawler): add content change detection and timestamp management

yesidc:
## Summary
This PR introduces a  new feature for cache management - content change detection. Users can now specify whether they want to check if web content has changed before using a cached version.

When `check_content_changed` is enabled in `CrawlerRunConfig the crawler will use a  multi-tiered approach to determine if content has changed:

1. First, it checks cache-control headers for max-age directives
2. Then, it makes a conditional HEAD request using ETags and Last-Modified headers
3. Finally, it falls back to a configurable default TTL

Previously, the caching system would simply retrieve cached content without checking if the underlying web content had changed. This meant that users might receive stale data without knowing it.

With this new feature, users can now ensure they're always working with fresh content while still benefiting from caching performance. If the content has changed, the crawler automatically invalidates the cache, fetches fresh content, and updates the database.

The implementation is efficient, using low-latency HEAD requests and respecting standard HTTP caching mechanisms, minimizing unnecessary full page retrievals.

This enhancement provides significant value for applications requiring both performance and data freshness, and gives users fine-grained control over their caching strategy.

1.  `async_configs.py` : added the following parameters to enable content change detections:
- `check_content_changed (bool)`: If True, check if the content has changed before processing. Default: False.
                                      
-  `head_request_timeout (float):` Timeout in seconds for the initial HEAD request. Default: 3.0.
                                      
 -  `default_cache_ttl_seconds (int):`  Default time-to-live for cached responses in seconds. Default: None.

2. `async_webcrawler.py`:  added `_check_content_changed`  method to check if the cached content has changed.  Also modified `arun` accordigly. 
3.  `async_database`: Added a new column `timestamp` and a method `adelete_data_point` to delete urls from db. 

## How Has This Been Tested?
Basic Testing:

1. Crawl a URL with `check_content_changed=False` (default behavior)
  - Crawl the same URL again - it should use the cached version
  - Modify the website content (or wait for it to change naturally)
  - Crawl with `check_content_changed=True` - it should detect changes and fetch fresh content

2. Find a URL that returns proper caching headers (Cache-Control, ETag, Last-Modified)
  - Crawl with check_content_changed=True
  - Crawl again immediately - it should use cache (fresh according to headers)

### DIFF
--- a/crawl4ai/async_configs.py
+++ b/crawl4ai/async_configs.py
@@ -756,6 +756,12 @@ class CrawlerRunConfig():
                                Default: False.
         shared_data (dict or None): Shared data to be passed between hooks.
                                      Default: None.
+        check_content_changed (bool): If True, check if the content has changed before processing.
+                                      Default: False.
+        head_request_timeout (float): Timeout in seconds for the initial HEAD request.
+                                      Default: 3.0.
+        default_cache_ttl_seconds (int): Default time-to-live for cached responses in seconds.
+                                          Default: None.
 
         # Page Navigation and Timing Parameters
         wait_until (str): The condition to wait for when navigating, e.g. "domcontentloaded".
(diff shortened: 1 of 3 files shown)


unclecode:
Thanks for the effort, but this is too large in scope for us to review and merge. If you'd like to contribute, consider breaking it into smaller, focused PRs that are easier to review.

PR #1207, opened 2025-06-10
Fix/streaming error handling

moncapitaine:
## Summary
If you stream an arun crawl, e.g. 

```python
            arunIterator = await crawler.arun(url, config=config)
            while True:
                result = await arunIterator.__anext__()
                try:
                    print(f"Start Processing crawler result")
                    yield "\n---BEGIN CRAWLER RESULT---\n"
                    print(f"Processing URL: {result.url}")
                    yield json.dumps({"url": result.url, "title": result.metadata.get("title")})
                    yield "\n---BEGIN MARKDOWN---\n"
                    print(f"Processing Markdown: {result.url}")
                    yield result.markdown
                except Exception as e:
                    # On any error, yield an error JSON to keep stream alive
                    print(f"Error processing URL {url}: {str(e)}")
                    yield "\n---ERROR---\n"
                    yield json.dumps({"error": str(e)})
                    yield "\n---END ERROR---\n"
                finally:
                    print(f"Finished processing URL")
                    yield "\n---END CRAWLER RESULT---\n"
```

with a deep crawling strategy like 

```python
        config = CrawlerRunConfig(
            stream=True,
            deep_crawl_strategy=BFSDeepCrawlStrategy(
                max_depth=max_depth,
                max_pages=max_pages, 
                include_external=False
            ),
            scraping_strategy=LXMLWebScrapingStrategy(),
            verbose=True,
        )

```

The run will not continue on the first page that causes an error, e.g. navigate, not-found (dead links in the target) because the iterator will stop because of the raised error from crawl4ai.

I changed
async_crawler_strategy.py

to not raise an error on arun in case of streaming.

## How Has This Been Tested?

I tested my code example to see if it continues crawling on various errors on badly maintained public web pages.

## Checklist:

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/docs/md_v2/core/simple-crawling.md
+++ b/docs/md_v2/core/simple-crawling.md
@@ -31,16 +31,9 @@ if __name__ == "__main__":
 The `arun()` method returns a `CrawlResult` object with several useful properties. Here's a quick overview (see [CrawlResult](../api/crawl-result.md) for complete details):
 
 ```python
-config = CrawlerRunConfig(
-    markdown_generator=DefaultMarkdownGenerator(
-        content_filter=PruningContentFilter(threshold=0.6),
-        options={"ignore_links": True}
-    )
-)
-
 result = await crawler.arun(
     url="https://example.com",
-    config=config
+    config=CrawlerRunConfig(fit_markdown=True)
 )
 
 # Different content formats


unclecode:
Thank you for identifying this streaming error handling concern. After reviewing the current codebase, the issue you raised is now handled by the caller in async_webcrawler.py. The arun() method catches exceptions from the crawl strategy (line 519), retries with other proxies if available, and when all attempts fail it builds a minimal failed CrawlResult (lines 620-627) instead of crashing. This means the streaming pipeline (arun_many with stream=True) gracefully continues to the next URL even when one URL throws an exception. So the raise in async_crawler_strategy.py is the correct behavior - it lets the caller's retry and fallback logic do its job. Appreciate you raising this, and thanks for contributing\!

Issue #123, opened 2024-10-02
Timeout setting


openainext:
Timeout setting

unclecode:
@openainext Hi, can you explain more on this? Thx

PR #1211, opened 2025-06-11
Fix: safely create a new page if no page exists in persistent context

### LINKED ISSUE #1198: [Bug]: use_persistent_context=True causes crawler to crash in concurrent mode due to empty context.pages access
### crawl4ai version

0.6.3

### Expected Behavior

If no page exists in the persistent context, a new page should be created via:

`page = await context.new_page()`

### Current Behavior

When using BrowserConfig(use_persistent_context=True, user_data_dir=...), the crawler crashes during concurrent runs when context.pages is empty.

This happens inside BrowserManager.get_page() when attempting to access context.pages[0] directly without checking if the list is empty:

This becomes especially problematic during concurrent execution (arun_many) where multiple URLs are processed in parallel, and a new page is expected but not guaranteed to exist.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python
In the browser_manager.py:
        if self.config.use_managed_browser:
            context = self.default_context
            pages = context.pages
            page = next((p for p in pages if p.url == crawlerRunConfig.url), None)
            if not page:
                page = context.pages[0] # await context.new_page()
If I change the last row into:
  page = context.pages[0] if context.pages else await context.new_page()

It works in my case.
```

### OS

Windows

### Python version

3.13.3

### Browser

_No response_

### Browser version

_No response_

_No response_

Praneeth1-O-1:
Fixes a crash that occurs when using use_managed_browser=True and the context.pages list is empty during concurrent execution.
Previously, context.pages[0] was accessed directly without checking if the list was empty, which led to a list index out of range or context-closed error.
This fix ensures that a new page is created if no pages exist.

Fixes #1198

eg: `Fixes #123` (Tag GitHub issue numbers in this format, so it automatically links the issues with your PR)

crawl4ai/browser_manager.py
→ Updated get_page() method to safely handle the case when no pages exist in a context by checking if context.pages is empty and creating a new page when needed.

Ran the crawler with use_persistent_context=True and multiple URLs using arun_many.

Confirmed that:

No crashes occurred when context.pages was empty.

New pages were created when needed.

Existing pages were reused when possible.

### DIFF
--- a/crawl4ai/browser_manager.py
+++ b/crawl4ai/browser_manager.py
@@ -964,7 +964,10 @@ class BrowserManager:
             pages = context.pages
             page = next((p for p in pages if p.url == crawlerRunConfig.url), None)
             if not page:
-                page = context.pages[0] # await context.new_page()
+                if pages:
+                        page = pages[0]
+                else:
+                    page = await context.new_page()        
         else:
             # Otherwise, check if we have an existing context for this config
             config_signature = self._make_config_signature(crawlerRunConfig)


ntohidi:
@Praneeth1-O-1 pls provide a test script. I'll take a look. thanks

Praneeth1-O-1:
> @Praneeth1-O-1 pls provide a test script. I'll take a look. thanks

I have added an example test script in tests/test_browser.py, take a look and get back to me if you face any issues.
=> Added `tests/test_browser.py` to test persistent context page creation. Tests 1, 2, 4, 5 passed, validating the fix.

JamesN-dev:
I also hit this exact same bug today while building a documentation scraper that processes multiple URLs in sequence.

This fix is exactly right. This explains why browser_mode="builtin" (which internally forces use_managed_browser=True) was hitting this issue even without explicit session management.

I temporarily worked around it by removing browser_mode="builtin" from BrowserConfig, this confirmed the managed browser context reuse was the root cause.

Your testing with arun_many matches my use case perfectly. Any timeline on getting this merged? This is blocking multi-URL batch processing workflows.

unclecode:

PR #1308, opened 2025-07-16
Fixed the error that the variable css_selector was originally assigned to list

dominicx:
## Summary
In CrawlerRunConfig, the css_selector is string, and will be split. but these files in doc css_selector is assigned an initial value of list type.

eg: quickstart.py - To update the example as per new changes
docs\md_v2\assets\llm.txt\txt\config_objects.txt - To change css_selector's value
docs\md_v2\assets\llm.txt\txt\llms-full.txt  - To change css_selector's value
docs\md_v2\assets\llm.txt\txt\llms-full-v0.1.1.txt  - To change css_selector's value


unclecode:
Thank you! The `css_selector` type has been corrected from list to string in the documentation examples and merged into `develop`. Appreciate the contribution!

PR #1337, opened 2025-07-25
fix: invert URL scores in BestFirstCrawlingStrategy priority queue to…

### LINKED ISSUE #1253: [Bug]: BestFirstCrawlingStrategy: Priority inversion and premature link truncation break best-first semantics
### crawl4ai version

0.6.3

### Expected Behavior

1) Priority ordering
According to the [documentation](https://docs.crawl4ai.com/core/deep-crawling/), BestFirstCrawlingStrategy should visit higher-scoring pages first:

“This strategy prioritizes URLs based on their score, ensuring that higher-value pages are crawled first.”

So, a URL scored at 0.9 should be crawled before a URL scored at 0.2.

2) Link discovery
All valid discovered links should be scored and added to the queue. The max_pages limit should stop the total number of pages fetched, not limit how many URLs are discovered before scoring.

### Current Behavior

1) Priority inversion
The implementation pushes the raw score to an asyncio.PriorityQueue, which is a min-heap.

This causes lower scores to be crawled first, which is the opposite of the intended best-first traversal.

2) Premature truncation
In link_discovery, the number of discovered valid links is capped by the remaining max_pages before it even gets to the scorer, applying an arbitrary truncation.

This means links are discarded before scoring, so relevant pages may never be added to the queue. Truncation is not necessary as the crawler will anyway stop once max pages is reached, and doing it this way ensures highest score pages are never lost.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
Any inputs when using the BestFirstCrawlStrategy
```

### Steps to Reproduce

```bash
Use BestFirstCrawlingStrategy with max_pages set low (e.g. 2–3).

Use a keyword scorer that produces different scores for known paths (e.g. 1 for /about, 0 for /privacy).

Crawl a page that has several internal links.

Observe that the lower-scoring pages are crawled first and some valid pages are never scored at all.
```

### Code snippets

```python
bff_strategy.py

// 1. Min heap implementation causes lowest scores to be dequeued first
await queue.put((new_score, new_depth, new_url, new_parent))

// 2. Truncation occurs before scoring
if len(valid_links) > remaining_capacity:
    valid_links = valid_links[:remaining_capacity]
```

### OS

Linux

### Python version

3.12.3

### Browser

_No response_

### Browser version

_No response_

_No response_

piyush-gpt:
Fixes the crawling priority logic in BestFirstCrawlingStrategy by inverting URL scores before inserting them into the priority queue. The issue was that the strategy uses a min-priority queue which caused URLs with lower scores to be crawled first, contradicting the intended max-priority behavior where higher scored URLs should be crawled earlier.

This change negates the scores when enqueuing URLs, and negates them back when yielding crawl results. This aligns crawl order with scorer priorities and respects documented behavior for prioritized crawling..

Fixes #1253 

crawl4ai/deep_crawling/bff_strategy.py :
Fixed the priority inversion by inserting URLs with negative scores into the priority queue and restoring positive scores on crawl result metadata.

## How Has This Been Tested?
Manually tested with custom WarrantyPriorityScorer scoring URLs with varying positive scores to verify that URLs with higher scores are crawled before lower scored URLs.
Verified crawling logs show that URLs are processed in descending order of their scorer rank.
Confirmed crawl stops on reaching max_pages as before, with no regression in other crawl behaviors.

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/crawl4ai/deep_crawling/bff_strategy.py
+++ b/crawl4ai/deep_crawling/bff_strategy.py
@@ -187,7 +187,7 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
                 result.metadata = result.metadata or {}
                 result.metadata["depth"] = depth
                 result.metadata["parent_url"] = parent_url
-                result.metadata["score"] = score
+                result.metadata["score"] = -score #original score as metadata
                 
                 # Count only successful crawls toward max_pages limit
                 if result.success:
@@ -208,7 +208,7 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
                     for new_url, new_parent in new_links:
                         new_depth = depths.get(new_url, depth + 1)
                         new_score = self.url_scorer.score(new_url) if self.url_scorer else 0
-                        await queue.put((new_score, new_depth, new_url, new_parent))
+                        await queue.put((-new_score, new_depth, new_url, new_parent)) # negative of score so as to prioritize higher positive score from min heap
 
         # End of crawl.
 


ntohidi:
Thank you for your contribution💜. We will add it to our next release.

Issue #1335, opened 2025-07-25
[Bug]: Improve docker error handling


nuo-wen-lei:
### crawl4ai version

0.6.3

### Expected Behavior

`
class CrawlResult(BaseModel):
    url: str
    html: str
    success: bool
    cleaned_html: Optional[str] = None
    media: Dict[str, List[Dict]] = {}
    links: Dict[str, List[Dict]] = {}
    downloaded_files: Optional[List[str]] = None
    screenshot: Optional[str] = None
    pdf : Optional[bytes] = None
    mhtml: Optional[str] = None
    markdown: Optional[Union[str, MarkdownGenerationResult]] = None
    extracted_content: Optional[str] = None
    metadata: Optional[dict] = None
    error_message: Optional[str] = None
    session_id: Optional[str] = None
    response_headers: Optional[dict] = None
    status_code: Optional[int] = None
    ssl_certificate: Optional[SSLCertificate] = None
    class Config:
        arbitrary_types_allowed = True
`

This is the expected results. When there is an error such as visiting a non-existent webpage or some crawler error, I expect fields such as `status_code` and `error_message` to be populated.

### Current Behavior

This is the docker result output when crawling a non-existent webpage with `/crawl` or `/crawl/job`. No fields are populated, however on the crawl4ai docker output, it does print out a useful error message, which I would like to see in the docker output. (Check error logs section)

<img width="1277" height="127" alt="Image" src="https://github.com/user-attachments/assets/363d8d85-15ae-42cc-80d1-1f7f317f6654" />

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
1. Start up a crawl4ai docker container
2. Visit /docs
3. Try to scrape a non-existent webpage

{
  "urls": [
"http://foobarhlkajslkjf.com"
],
  "browser_config": {
    
  },
  "crawler_config": {
  }}

4. Check results
```

### Code snippets

```python

```

### OS

macOS

### Python version

3.12 (using crawl4ai docker)

### Error logs & Screenshots (if applicable)

Error log contains expected error message

```
crawl4ai  | [ERROR]... × http://foobarhlkajslkjf.com                        | Error: 
crawl4ai  | Unexpected error in _crawl_web at line 744 in _crawl_web 
crawl4ai  | (../usr/local/lib/python3.12/site-packages/crawl4ai/async_crawler_strategy.py):
[...]

PR #1441, opened 2025-08-26
Improve docker error handling

SohamKukreti:
This PR does the following:
- Return comprehensive error messages along with status codes for api internal errors.
- Fix fit_html property serialization issue in both /crawl and /crawl/stream endpoints
- Add sanitization to ensure fit_html is always JSON-serializable (string or None)
- Add comprehensive error handling test suite.

Addresses #1335

`deploy/docker/api.py`: To fix fit_html causing serialization issue when scraping incorrect URL.
`deploy/docker/server.py`: To catch errors and exceptions in the api server.
`tests/test_server_request.py`: To add tests for error handling.

## How Has This Been Tested?
Ran docker tests and tested manually with correct and incorrect URLs.

## Checklist:

### DIFF
--- a/deploy/docker/api.py
+++ b/deploy/docker/api.py
@@ -413,6 +413,9 @@ async def stream_results(crawler: AsyncWebCrawler, results_gen: AsyncGenerator) 
                 server_memory_mb = _get_memory_mb()
                 result_dict = result.model_dump()
                 result_dict['server_memory_mb'] = server_memory_mb
+                # Ensure fit_html is JSON-serializable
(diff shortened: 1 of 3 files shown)


deploy/docker/api.py
jizhilong, line 1: Instead of hacking around the type of `result.fit_html`, I suggest to just remove the `crawl4ai.models.CrawlResult.fit_html` field,  since this field is overriden by the `@property def fit_html`, and the `raise` statement within the body of `@property def fit_html` suggest that such override is by design.

PR #1444, opened 2025-08-27
Fix: remove  hardcoded openai llm in adaptive_crawler

### LINKED ISSUE #1291: [Bug]: Crawl4AI always defaults to OpenAI instead of using GROQ_API_KEY from .llm.env
### crawl4ai version

0.6.1

### Expected Behavior

When I define GROQ_API_KEY in the .llm.env file and run the container with --env-file .llm.env, I expect Crawl4AI to use Groq as the LLM provider (llama3) instead of OpenAI.

### Current Behavior

Even though I correctly set GROQ_API_KEY=sk-xxx in the .llm.env file, Crawl4AI still logs that it is using openai as the provider with model gpt-4o-mini. This indicates that Groq is not being picked up or used at all.

Logs show:

LiteLLM completion() model= gpt-4o-mini; provider = openai

I confirmed that the .llm.env file is mounted correctly and the container is started using:

docker run -d -p 11235:11235 \
  --name crawl4ai \
  --env-file .llm.env \
  --shm-size=1g \
  unclecode/crawl4ai:latest

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python

```

### OS

Windows WSL

### Python version

3.10.12

### Browser

_No response_

### Browser version

_No response_

_No response_

ShaneM123:
## Summary
Please include a summary of the change and/or which issues are fixed.

eg: `Fixes #1291` (Tag GitHub issue numbers in this format, so it automatically links the issues with your PR)

adaptive_crawler.py - remove hardcoded llm from  adaptive_crawler.
The hardcoded openai llm was preventing users from selecting an llm of their choice. use a env var instead to allow use of any supported llm

## How Has This Been Tested?
Ran changes locally using the adaptive_crawler feature in a python script. After this change i was able to set the llm to any litellm whereas before it was hardcoded.

## Checklist:


ShaneM123:
This is a quick fix patch, honestly the use of configs and env vars would needs  a larger refactoring for a clean solution, having env vars for every single API_KEY type is impractical. 

Having multiple possible configs, one for embedded and one for not, adds to the confusion and may be why this was missed. There is no need to hard code open_ai models anywhere in the code except perhaps in tests.

I would recommend to follow this up with another PR for a future release

ntohidi:
> This is a quick fix patch, honestly the use of configs and env vars would needs a larger refactoring for a clean solution, having env vars for every single API_KEY type is impractical.
> 
> Having multiple possible configs, one for embedded and one for not, adds to the confusion and may be why this was missed. There is no need to hard code open_ai models anywhere in the code except perhaps in tests.
> 
> I would recommend to follow this up with another PR for a future release

I agree that this change is necessary. We have decided not to pass the llm_config as a dictionary to the adaptive crawler. Instead, we will pass the LLMConfig object, allowing users to set the LLM provider and the API key from that object.
I will create a new issue for this, and we will work on it.

ShaneM123:
Okay, I will close this PR then, thanks.

PR #1450, opened 2025-08-28
Fix: LLM extraction fails when content is in alternative response fields (content_reason, provider_specific_fields.refusal)

### LINKED ISSUE #1449: [Bug]: LLM extraction fails when content is in alternative response fields (content_reason, provider_specific_fields.refusal)
### crawl4ai version

0.7.4

### Expected Behavior

The LLM extraction strategy should successfully extract content from LLM responses regardless of which field contains the actual content. The system should check multiple possible fields (`content`, `content_reason`, `provider_specific_fields.refusal`) to retrieve the response content.

### Current Behavior

The `LLMExtractionStrategy` only checks the `content` field for response content. When LLM models return content in alternative fields like `content_reason` or `provider_specific_fields.refusal`, the extraction fails with the error: `"Crawl returning 'str' object has no attribute 'choices'"`

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- Testing with `o1-mini` model: content sometimes appears in `provider_specific_fields.refusal` instead of `content`
- Testing with `DeepSeek-V3` model: content sometimes appears in `content_reason` instead of `content`
```

### Steps to Reproduce

```bash
1. Configure crawl4ai with LLM extraction strategy using `o1-mini` or `DeepSeek-V3` models
2. Perform crawling operations where the LLM response structure varies
3. Observe that when `content` field is `None` but content exists in other fields, extraction fails
```

### Code snippets

```python
The issue occurs in the LLM extraction logic where only the `content` field is checked. A more robust approach would implement a fallback mechanism to check alternative fields when `content` is empty or `None`.
```

### OS

All

### Python version

3.12

### Browser

_No response_

### Browser version

_No response_

Crawl returning 'str' object has no attribute 'choices

rbushri:
## Summary
Fixes LLM extraction failures when content is returned in alternative response fields instead of the main `content` field. Some LLM providers (like o1-mini and DeepSeek-V3) occasionally return content in `content_reason` or `provider_specific_fields.refusal` fields, causing extraction to fail with "str object has no attribute 'choices'" error.

This PR implements a fallback mechanism in the LLM extraction strategy to check multiple response fields when the primary `content` field is empty or None.

Fixes #1449

- `crawl4ai/extraction_strategy.py` - Updated LLMExtractionStrategy to implement content extraction fallback mechanism
- `tests/test_extract_response_content.py` - Added unit tests to verify fallback behavior with different response structures

## How Has This Been Tested?
- Added unit tests covering scenarios where content appears in `content_reason` and `provider_specific_fields.refusal`
- Tested with o1-mini model responses where content was in `provider_specific_fields.refusal`
- Tested with DeepSeek-V3 model responses where content was in `content_reason`
- Verified existing functionality remains unchanged when content is in the primary `content` field
- All existing unit tests pass with the changes

## Checklist:

- [] I have made corresponding changes to the documentation

### DIFF
--- a/crawl4ai/extraction_strategy.py
+++ b/crawl4ai/extraction_strategy.py
@@ -46,6 +46,32 @@ from bs4 import BeautifulSoup
 from lxml import html, etree
 
 
+def extract_response_content(response):
+    """
+    Extract content from LLM response with fallback options.
+
+    Args:
+        response: The LLM response object
+
+    Returns:
+        str: The content from the response, trying multiple sources in order:
+             1. message.content
+             2. message.reasoning_content
+             3. message.provider_specific_fields.refusal
+             4. None if all fail
+    """
+    msg = response.choices[0].message
+
+    content = (
+        msg.content
+        or getattr(msg, "reasoning_content", None)
+        or (getattr(msg, "provider_specific_fields", {}) or {}).get("refusal")
+        or None
+    )
+
+    return content
+
+
 class ExtractionStrategy(ABC):
     """
     Abstract base class for all extraction strategies.
@@ -656,7 +682,7 @@ class LLMExtractionStrategy(ExtractionStrategy):
             self.total_usage.total_tokens += usage.total_tokens
 
             try:
-                content = response.choices[0].message.content
+                content = extract_response_content(response)
                 blocks = None
 
                 if self.force_json_response:
@@ -680,7 +706,7 @@ class LLMExtractionStrategy(ExtractionStrategy):
                     block["error"] = False
             except Exception:
                 parsed, unparsed = split_and_parse_json_objects(
-                    response.choices[0].message.content
+                    extract_response_content(response)
                 )
                 blocks = parsed
                 if unparsed:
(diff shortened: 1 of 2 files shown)


unclecode:
Thanks for the detailed report and PR @rbushri.

We investigated this thoroughly and concluded this isn't a bug in crawl4ai's extraction logic. Here's what we found:

**litellm already handles response field normalization.** When you use any provider through litellm (which crawl4ai uses internally), the actual answer is always placed in `response.choices[0].message.content`. For reasoning models like DeepSeek-R1, litellm separates the chain-of-thought (`reasoning_content`) from the final answer (`content`). Our code correctly reads `content`.

**The proposed fallback would produce wrong results.** `reasoning_content` contains the model's internal thinking process (e.g., "Let me analyze this HTML..."), not the extraction output. Falling back to it when `content` is None would inject reasoning monologue into the extraction pipeline. Similarly, `refusal` means the model refused the request — treating that as valid extracted data would silently hide errors.

**The error you're seeing (`'str' object has no attribute 'choices'`) suggests the response object itself isn't a valid ModelResponse.** This points to either:
- An older litellm version (we require >=1.53.1, current is 1.81.5) — newer versions have better provider handling
- A provider-specific edge case where litellm didn't parse the response correctly

**What we'd suggest:**
1. Update litellm to the latest version (`pip install -U litellm`)
2. If the issue persists, open a litellm issue with the exact provider/model and a minimal reproduction — they're responsive and this is their domain

We also couldn't find a `content_reason` field in litellm's Message model, which the issue mentions for DeepSeek-V3.

Closing this PR, but appreciate you taking the time to investigate and write tests. If you can reproduce this with the latest litellm, please reopen the issue with a minimal example and we'll revisit.

PR #1640, opened 2025-11-25
fix: prevent memory leak by closing unused context

### LINKED ISSUE #943: [Bug]: After the crawl4ai container runs for a few minutes, a large number of chrome processes remain, consuming CPU and memory. I hope to solve this problem.
### crawl4ai version

docker image tag: unclecode/crawl4ai:basic-amd64

### Expected Behavior

When crawl4ai finishes crawling the web page, the used Chrome process should be released immediately to avoid a large number of Chrome processes remaining in the background.

Or is there an available container configuration that can achieve this goal?

### Current Behavior

Image

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
curl --location 'http://127.0.0.1:11235/crawl' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer 12345678' \
--data '{
    "urls": "test url here",
    "crawler_params": {
        "headless": true,
        "page_timeout": 15000,
        "remove_overlay_elements": true,
        "semaphore_count": 1
    },
    "extra": {
        "word_count_threshold": 20,
        "bypass_cache": true,
        "only_text": true,
        "process_iframes": false
    }
}'
```

### Steps to Reproduce

```bash
docker run -p 11235:11235 --env CRAWL4AI_API_TOKEN=12345678 --env MAX_CONCURRENT_TASKS=1 --name crawl4ai -m 4G --restart unless-stopped unclecode/crawl4ai:basic-amd64
```

### Code snippets

```python

```

### OS

Linux Docker

### Python version

3.10.15

### Browser

_No response_

### Browser version

_No response_

_No response_

Martichou:
## Summary
When scraping many URLs continuously, browser contexts accumulate in memory and are never cleaned up. The existing cleanup mechanism only runs when browsers go idle, which never happens under continuous load. This causes memory to grow unbounded until the process crashes or becomes unresponsive.

Fixes #943 

> Small note: I'm not used to python, I won't lie, Claude helped me a bit here, but I've checked what it did and tested it. So this is not just yet another AI slop :) 

- `browser_manager.py`: Add _context_refcounts tracking, cleanup_contexts(), and release_context() methods
- `async_crawler_strategy.py`: Release context ref in finally block after crawl
- `deploy/docker/api.py`: Trigger context cleanup after each request

## How Has This Been Tested?
This has been tested locally by running the following script and comparing the before/after memory usage with both the master version and the patched version through a docker compose.

The script simply perform 100 scrape with 8 concurrency and report the status code repartition:
https://gist.github.com/Martichou/27555055d130d1c65f6a8457fbeb2a22

Result of the test:

Unpatched version:
```
Baseline memory usage: 4.5%
End of first test run using unpatched version: 23.4%
End of second test run using unpatched version: 27.6%
End of third test run using unpatched version: 32.8%
```

Patched version:
```
Baseline memory usage: 5.7%
End of first test run using unpatched version: 11.2%
End of second test run using unpatched version: 12.3%
End of third test run using unpatched version: 13.4%
```

It may not have eliminated every leaks (1% gains between run for unknown reason), but closing the browser using the kill browser endpoint make the memory go back to 10%. 

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     lsof \
     redis-server \
     supervisor \
+    tini \
     && apt-get clean \ 
     && rm -rf /var/lib/apt/lists/*
 
@@ -200,7 +201,12 @@ EXPOSE 6379
 USER appuser
 
 # Set environment variables to ptoduction
-ENV PYTHON_ENV=production 
+ENV PYTHON_ENV=production
+
+# Use tini as init system to properly reap zombie processes
+# This is required for Playwright/Chromium which spawns many child processes
+# See: https://github.com/unclecode/crawl4ai/issues/1666
+ENTRYPOINT ["/usr/bin/tini", "--"]
 
 # Start the application using supervisord
 CMD ["supervisord", "-c", "supervisord.conf"]


unclecode:
Hi @Martichou — thanks for this PR and the detailed memory profiling data. Your diagnosis of the problem was spot on: browser contexts in `contexts_by_config` were never cleaned up under continuous load, causing unbounded memory growth.

We ended up implementing the fix directly on `develop` with a slightly different (but inspired by your work) approach that accounts for recent changes to `browser_manager.py` (global page tracking, CDP connection caching, etc.) that would have caused heavy merge conflicts with this PR. Here's what we landed:

### 1. Config signature shrink (the bigger win)
The root cause of excessive context creation was that `_make_config_signature()` was hashing **~60+ fields** from `CrawlerRunConfig`, but only **7 fields** actually affect the browser context (`proxy_config`, `locale`, `timezone_id`, `geolocation`, `override_navigator`, `simulate_user`, `magic`). Fields like `word_count_threshold`, `css_selector`, `screenshot`, `verbose`, etc. were producing phantom-different signatures for browser-identical contexts. We switched from a blacklist to a whitelist approach.

### 2. Refcount tracking + LRU eviction (your core idea)
Following your reference-counting concept, we added:
- `_context_refcounts` — tracks active crawls per context (increment in `get_page()` under lock, decrement in `release_page_with_context()` under lock)
- `_context_last_used` — monotonic timestamps for LRU ordering
- `_evict_lru_context_locked()` — when contexts exceed `_max_contexts` (default 20), evicts the oldest context with `refcount == 0`
- Eviction only targets idle contexts (never evicts mid-crawl), and context close happens outside the lock to avoid blocking

### 3. Storage state path leak fix
Also fixed a separate leak where the `storage_state` code path created a temporary context for cloning that was never closed.

### Test coverage
Added 20 tests (8 signature, 5 eviction logic, 7 real-browser integration) — all passing. The real-browser tests verify: same context reused for non-context config changes, refcounts drop to 0 after crawl, LRU eviction caps contexts, concurrent crawls track correctly.

Your PR was the catalyst for this investigation — really appreciate the contribution and the benchmarking script. Closing in favor of the direct implementation on develop.

unclecode:
@Martichou Thanks for the detailed memory analysis — really helpful.

I've pushed additional fixes to the `develop` branch that build on top of your findings:

1. **Memory-saving Chrome flags** — disabled unused Chrome features (OptimizationHints, MediaRouter, component updates, domain reliability) by default, plus an opt-in `memory_saving_mode` for aggressive cache discard and V8 heap cap (512MB)

2. **Browser recycling** — new `max_pages_before_recycle` config option that automatically restarts the browser process after N pages to reclaim leaked Chromium memory. Blocks new crawls during recycle, wakes them when done. Recommended: 500-1000 for long-running crawlers

3. **CDP session leak fix** — `cdp.detach()` after viewport adjustment to prevent orphaned DevTools sessions from accumulating

4. **Session kill fix** — context is now only closed when its refcount drops to 0, preventing use-after-close errors for shared contexts

Usage:
```python
config = BrowserConfig(
    memory_saving_mode=True,           # aggressive cache + V8 heap cap
    max_pages_before_recycle=500,      # recycle browser every 500 pages
)
```

Could you pull the `develop` branch and re-run your memory tests to see how these changes affect the growth pattern? Especially interested in whether the recycling + CDP detach combo brings the curve back down.

Martichou:
@unclecode Hey, thanks for following-up!

I've just tested the new max_pages_before_recycle. A couple of things to say; I think this is exactly what's needed, so kudo for that! But this should be more aggressive, when going through my backlog of links to scrape, the Crawl4AI instance perform more than 20 concurrent requests, which means the recycle will never be performed due to the various check for no-inflight crawl.

In my testing, doing crawl one by one manually allow the recycle to trigger, but with 20 concurrent there's no chance.
I've set the max `max_pages_before_recycle ` to 5, and already am at more than 189 hits without recycle.


What would you think of when a recycle should happens, we either:
- block all new incoming until there's no more in-flight to allow the recycle to trigger (with a max timeout and force kill if needed).
- instead of blocking new incoming, we force re-create a new instance of the browser, to be used from now on for this config and this would let the previous one be recycled/killed.
- keep the recycle as is, but have a new option be "new browser after X", or this can be simply a params in the browser_config to specify the "version" so the caller can decide when he want a new instance to be created? And this version would be used in the hash of the config to see which browser to use?


Lemme know what you think, and thank you for your help here :)

unclecode:
@Martichou Thanks for testing and the great feedback! You were right — the old approach never triggered under sustained concurrent load because it waited for `total_active == 0`.

I've pushed a completely new approach based on your suggestion (option 2 — spin up a new browser while old one drains):

### How it works now:

Instead of waiting for a quiet moment, we **bump a version number** and let old browsers drain naturally:

1. **`_browser_version`** is now part of the config signature
2. When threshold is hit → bump version → reset counter
3. New requests automatically get new contexts (different version = different signature)
4. Old contexts stay in `_pending_cleanup` and drain naturally
5. When old context's refcount drops to 0, it gets closed
6. Safety cap: max 3 old browsers draining at once (blocks if exceeded)

**No blocking, no waiting** — old and new browsers coexist briefly during transitions. This means recycling now works under any load pattern.

### Usage (unchanged):
```python
config = BrowserConfig(
    memory_saving_mode=True,
    max_pages_before_recycle=500,
)
```

Could you pull `develop` again and re-run your memory tests? This time with 20+ concurrent crawls, you should see the version bumps happening in the logs and memory being reclaimed as old contexts drain.

Martichou:
Hey @unclecode, thanks for the follow-up again.

I've just test the develop branch (with the newest changes) and it's not behaving as expected. I'm not sure why, I can't really make sense of all the logs but from what I see and understand is that the old contexts never get closed because the request seems to never complete(?).
_Maybe unrelated but there's quite a lot of error too in the logs._

From what I see in the UI: more than 90+ active request, with 900+ seconds of being active; when usually it's between 10s and 90s, no more (timeout of 60s and hard timeout of 2 minutes). My system is doing at most 24 concurrents requests.

I guess something's wrong in the cleanup / count or something? I wish I could help more here, but this is kinda beyond my Python understanding right now (and inner working of Crawl4AI).

```
Helped with AI (disclosing cause maybe stupid):

Can the release_page_with_context runs before _maybe_bump_browser_version adds the sig to _pending_cleanup? Because then the sig enters pending with refcount already at 0, and no future release will trigger cleanup?

Suggested fix by Claude:
Fix: After adding sigs to _pending_cleanup, immediately check if any already have refcount 0 and clean them up right there.
```

I'll let you decide and see what you want to do, but I highly appreciate your time and your effort in solving this issues once and for all! Kudo for that 🫶 

[hanged.log](https://github.com/user-attachments/files/25108182/hanged.log)

unclecode:
Ok this is interesting, at this point, I need to be able reproduce the issue and not anymore speculation. Can you share with me a python script to show how you are using the library for this crawl? I am not asking the exact part of ur codebase. I need you isolated that part of your code (without change), and put in a separate file where running the file reproduce the issue. Then I use that exact code as test case. At this point I need this. Let me know if this is doable?

Martichou:
> I need you isolated that part of your code (without change), and put in a separate file where running the file reproduce the issue. Then I use that exact code as test case. At this point I need this. Let me know if this is doable? 

Sure this is doable. I'll do that over the weekend because this will take some time that I don't have during the week. I'll ping you here once it's ready, thanks!

Martichou:
Hey @unclecode, I took some time to pack a small, minimal, reproducible case. 

In my testing on a Mac with an M4 chip and 16GB of RAM, the problem occurs in 100% of the cases. Typically, after around 80-130 scrapes, it stops processing; requests hang for an unknown reason, and nothing advances or gets killed.

docker-compose + python test script: https://gist.github.com/Martichou/85e48e368eae81cb6892ca03a5c63fb3

If you need more help, feel free to ping me. Big thanks, I really appreciate your effort!

unclecode:
Appreciate @Martichou I will work on this soon, within this week or weekend.

Martichou:
Hey @unclecode, have you had any luck reproducing the issue locally? If you did not have time yet it's fine, just wanted to know if you faced any issue reproducing this on your side as I think this is still quite an important issue present.

I still have my pipeline with more than 350k+ sources running constantly since yesterday (I currently have a cronjob to restart the docker every 10 minutes).

Thanks for your time :)

unclecode:
@Martichou today and tomorrow I will check for sure, btw this part `I still have my pipeline with more than 350k+ sources running constantly ` is 🔥❤️‍🔥🔥❤️‍🔥 anyway I will update you soon

unclecode:
Hey @Martichou - I've identified and fixed the root causes. Your reproduction script and the race condition you suspected were both spot on.

## Three bugs found and fixed

**Bug 1 - Race condition in `_maybe_bump_browser_version()`**: When a version bump happened, ALL context signatures were added to `_pending_cleanup`, including ones with refcount 0. Since no future release would ever fire for an idle sig, it stayed stuck in pending forever. Fix: idle sigs (refcount 0) are now cleaned up immediately during the bump instead of going into pending.

**Bug 2 - Finally block leaking refcounts**: The `_crawl_web` finally block accessed `page.context.browser.contexts` before calling `release_page_with_context()`. If the browser crashed (OOM in Docker), that line throws and the refcount is never decremented - leaked permanently. Fix: `release_page_with_context()` is now the very first call in the finally block, in its own try/except.

**Bug 3 - Safety cap deadlock**: When 3+ stuck entries accumulated in `_pending_cleanup` (from bugs 1 & 2), `get_page()` blocked forever with no timeout. All workers pile up, no releases happen, permanent deadlock. This is the "90+ active requests, 900+ seconds" you saw. Fix: 30-second timeout on the wait, after which stuck entries (refcount 0) are force-cleaned.

## Please test

All fixes are on `develop`. Could you:

1. Pull `develop` branch
2. Run the regression test I added - all 17 checks should pass:
   https://github.com/unclecode/crawl4ai/blob/develop/tests/browser/test_repro_1640.py
3. Then re-run your docker-compose + crawl4ai-leak-test.py and let me know how it goes

Thanks!

Martichou:
Hey @unclecode, thank for the time you took to look into this again :)

Sadly I'm not coming with great news.. I've just tested your new changes and it still hanged :/ I've ran the same setup, docker-compose from above + python script to test and it will still stays "hanged" on "Waiting for old browser to drain (pending: 3)" with no new scrape happening and those present in the list ending up being 300s+.

During my little deep dive (trying to help you), I noticed some error related to add_init_script in my logs. Looking at the code I can see those are not (call to the add_init_script) in the try..except of the _crawl_web function but right after got a page. So is it possible that this single thing cause the refcount to never reach 0 and so the context is never released/cleared/recycled?

I tested the theory, added code, logs, and tried. And bingo, it's now working as intended :) I'll let you apply the fix I just said, but it make total sense to try..except everything after we get the page. I also have done more changes while trying to fix the issues, attached is the complete patch; feel free to look at it and pick something that can be interesting for you 👍  (to be clear, I did test with only the try thing, and it worked fine, so the rest is prolly garbage).

Hope this helps, we're near the end of this issue, I can smell it!

[hanged2.log](https://github.com/user-attachments/files/25424680/hanged2.log)
[leaks.patch](https://github.com/user-attachments/files/25424668/leaks.patch)

unclecode:
@Martichou Awesome, you nailed it, many thanks. From your patches, the main thing to complete this fix is moving the try up to cover lines 565-688. I noticed 1) created_at timestamp on pending entries and 2) _page_to_old_sig tracking. For now I don't add them, bcoz I believe there shouldn't be any more refcount leak paths. If it happens, then we add them. I'm updating the code base and applying the change. Meanwhile, you also test your side with only the "moving the try" and see if that's all you need. Yes, this was a very serious deep issue. Thanks to you, and now it's coming to an end. I really appreciate your contribution on it. We should work more together like this.

unclecode:
@Martichou The codebase is updated and I pushed it.

Martichou:
@unclecode Thanks for taking the time to resolve this issue. I tried my best not only to report it but also to provide insights that might help 😄 I’m glad it worked out. I appreciated working with you, have a great one! 🚀

unclecode:
@Martichou Honestly, this was one of the best collaborations I've had on Crawl4AI. I always enjoy such organic teamwork. Soon I'll check the other PRs you've opened. Have a great day.

PR #1676, opened 2025-12-23
feat(docker): add toggle to enable/disable permanent browser instance

mzyfree:
## Summary
This PR adds a new configuration option `use_permanent` to the Docker deployment mode. 

**Problem**: Currently, a permanent browser instance is always initialized at startup, which consumes roughly 200MB-300MB of memory regardless of actual usage. In resource-constrained environments or high-concurrency stress tests, this persistent instance contributes to memory accumulation and potential OOM (Out Of Memory) issues.

**Solution**: By introducing the `use_permanent` toggle, users can now choose to disable the persistent browser. When disabled, browser instances are created strictly on-demand and are fully managed by the `janitor` cleanup logic, ensuring better memory reclamation.

- **deploy/docker/server.py**: Modified the `lifespan` startup sequence to conditionally initialize the browser pool based on the `use_permanent` config. Added a fallback default of `True` to ensure zero breaking changes for existing users.
- **deploy/docker/config.yml**: Added the `use_permanent` flag under the `crawler.pool` section to make the feature discoverable and configurable.

## How Has This Been Tested?
1. **Default Behavior**: Verified that the permanent browser still starts as usual when the config is missing or set to `true`.
2. **On-Demand Mode**: Set `use_permanent: false` and verified the following:
    - Log confirmation: `🚫 Permanent browser instance is disabled by default to save memory.`
    - Baseline memory usage decreased significantly at container startup.
    - Successfully performed crawls; the system correctly fell back to creating on-demand instances in the `COLD_POOL`.
3. **Resource Reclamation**: Confirmed that on-demand browsers are correctly terminated by the `janitor` task after the `idle_ttl_sec` threshold is reached.

## Checklist:

- [x] I have made corresponding changes to the documentation (Config updated in `config.yml`)

### DIFF
--- a/deploy/docker/config.yml
+++ b/deploy/docker/config.yml
@@ -62,6 +62,7 @@ crawler:
   pool:
     max_pages: 40                          # ← GLOBAL_SEM permits
     idle_ttl_sec: 300                     # ← 30 min janitor cutoff
+    use_permanent: true                   # ← Toggle permanent browser instance
   browser:
     kwargs:
       headless: true
--- a/deploy/docker/server.py
+++ b/deploy/docker/server.py
@@ -122,10 +122,14 @@ async def lifespan(_: FastAPI):
     monitor_module.monitor_stats.start_persistence_worker()
 
     # Initialize browser pool
-    await init_permanent(BrowserConfig(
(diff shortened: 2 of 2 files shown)


mzyfree:
@unclecode please review the code.

PR #1697, opened 2026-01-08
fix: allow local embeddings in EmbeddingStrategy

### LINKED ISSUE #1658: [Bug]: Fix `EmbeddingStrategy._get_embedding_llm_config_dict` ignoring local embeddings
### crawl4ai version

0.7.7

### Expected Behavior

When `EmbeddingStrategy` is initialized without an explicit `embedding_llm_config` (or with `None`), I expect it to faithfully return `None` from `_get_embedding_llm_config_dict()`. This would allow the downstream `get_text_embeddings` utility to correctly switch to the local `sentence-transformers` implementation.

### Current Behavior

The method `_get_embedding_llm_config_dict` contains a fallback mechanism that **forces** an OpenAI configuration if `self.config` is present but has no LLM config.

**Relevant code in `adaptive_crawler.py` (lines 633-644):**
```python
    def _get_embedding_llm_config_dict(self) -> Dict:
        # ... check self.config ...
        
        # Fallback to default if no config provided
        return {
            'provider': 'openai/text-embedding-3-small',
            'api_token': os.getenv('OPENAI_API_KEY')
        }
```

Because this method *always* returns a dict, the logic in `utils.py`'s `get_text_embeddings` never reaches the `else` block for local embeddings:

```python
# utils.py
if llm_config is not None:
    # Uses LiteLLM (calls OpenAI)
else:
    # Uses sentence-transformers (UNREACHABLE via EmbeddingStrategy)
```

This effectively makes local embeddings dead code when using `AdaptiveCrawler`.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- **Config**: `AdaptiveConfig(strategy="embedding", embedding_llm_config=None)` (intended for local mode).
```

### Steps to Reproduce

```bash
1. Install `crawl4ai` with `sentence-transformers`.
2. Configure `AdaptiveCrawler` to use `"embedding"` strategy without providing `embedding_llm_config`.
3. Run `crawler.digest(...)`.
4. **Expected**: Uses local model.
5. **Actual**: Fails with OpenAI authentication error (or attempts to use OpenAI if key is in env), ignoring the user's intent to use local embeddings.
```

### Code snippets

```python

```

### OS

Windows 11

### Python version

3.11

### Browser

_No response_

### Browser version

_No response_

_No response_

majiayu000:
## Summary
Fixes #1658

The `_get_embedding_llm_config_dict` method was always returning a fallback OpenAI configuration when no `embedding_llm_config` was provided. This made local sentence-transformers embeddings unreachable when using `AdaptiveCrawler`.

- `crawl4ai/adaptive_crawler.py` - Changed `_get_embedding_llm_config_dict` to return `None` instead of a fallback dict, allowing `get_text_embeddings()` to use the local sentence-transformers path when the user intends to use local embeddings
- `tests/unit/test_embedding_strategy_local.py` - Added unit test to verify the fix

## How Has This Been Tested?
- Unit test verifies `_get_embedding_llm_config_dict()` returns `None` when no config provided
- Test uses AST parsing to avoid import dependency issues

## Checklist:

### DIFF
--- a/crawl4ai/adaptive_crawler.py
+++ b/crawl4ai/adaptive_crawler.py
@@ -630,18 +630,20 @@ class EmbeddingStrategy(CrawlStrategy):
         self._validation_embeddings_cache = None  # Cache validation query embeddings
         self._kb_similarity_threshold = 0.95  # Threshold for deduplication
     
-    def _get_embedding_llm_config_dict(self) -> Dict:
-        """Get embedding LLM config as dict with fallback to default."""
+    def _get_embedding_llm_config_dict(self) -> Optional[Dict]:
+        """Get embedding LLM config as dict.
+
+        Returns None when no explicit embedding config is provided,
+        allowing the caller to use local sentence-transformers instead
+        of falling back to OpenAI.
+        """
         if hasattr(self, 'config') and self.config:
             config_dict = self.config._embedding_llm_config_dict
             if config_dict:
                 return config_dict
-        
-        # Fallback to default if no config provided
-        return {
-            'provider': 'openai/text-embedding-3-small',
-            'api_token': os.getenv('OPENAI_API_KEY')
-        }
+
+        # Return None to allow local embeddings when no config provided
+        return None
         
     async def _get_embeddings(self, texts: List[str]) -> Any:
         """Get embeddings using configured method"""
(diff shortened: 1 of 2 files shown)


unclecode:
Closing — this is addressed by #1717 which has been merged. Thanks for the contribution though.

PR #1746, opened 2026-01-30
Fix: Avoid Common Crawl calls for sitemap-only URL seeding

ChiragBellara:
## Summary
Fixes the issue #1747.
Fixes an issue where **AsyncUrlSeeder** initialized the Common Crawl index even when **source="sitemap"** was specified. This caused unnecessary requests to Common Crawl and **ConnectTimeout** failures in environments where Common Crawl is unreachable. The change ensures Common Crawl is only initialized when explicitly requested and normalizes the **source** string to be case-insensitive. The ability to perform URL seeding using sitemap only should not depend on the availability of the Common Crawl indexes.

**async_url_seeder.py** – Updated source parsing logic to normalize and validate the source value, and to initialize the Common Crawl index only when "cc" is explicitly included. This prevents unintended Common Crawl calls during sitemap-only seeding.
Since the sources are matched using string matching, I've added normalization of the source tokens (case-insensitive, trimmed) to ensure values like "CC" or "cc + sitemap" or " cc" or " CC + sitemap  " are handled correctly.

## How Has This Been Tested?
- Added a regression test that simulates a Common Crawl outage by overriding **_latest_index()** to raise a **ConnectTimeout**.
- Verified that sitemap-only seeding no longer invokes Common Crawl and completes successfully.
- Confirmed that Common Crawl is still initialized when **source** includes **"cc"** (e.g. "cc" or "cc+sitemap").

## Checklist:

### DIFF
--- a/crawl4ai/async_url_seeder.py
+++ b/crawl4ai/async_url_seeder.py
@@ -400,17 +400,19 @@ class AsyncUrlSeeder:
         if self.logger and hasattr(self.logger, 'verbose') and config.verbose is not None:
             self.logger.verbose = config.verbose
 
-        # ensure we have the latest CC collection id
-        if self.index_id is None:
-            self.index_id = await self._latest_index()
-
         # Parse source parameter - split by '+' to get list of sources
-        sources = source.split('+')
+        sources = [s.strip().lower() for s in source.split("+") if s.strip()]
+
         valid_sources = {"cc", "sitemap"}
         for s in sources:
             if s not in valid_sources:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


unclecode:
Thanks for this fix @ChiragBellara — merged into develop. Clean and solid change. The sitemap-only seeding was unnecessarily hitting Common Crawl, and your guard fixes that properly. Also nice touch normalizing the source string with strip/lower.

I added the missing `@pytest.mark.asyncio` decorator on the test so it runs under pytest.

This will be in the next release (0.8.1). You'll be acknowledged in CONTRIBUTORS.md.

PR #1759, opened 2026-02-16
Fix AttributeError in FilterChain.add_filter (Tuple Immutability) #1753

### LINKED ISSUE #1753: [Bug]: `FilterChain.add_filter` incorrect type
### crawl4ai version

0.8.0

### Expected Behavior

You should be able to use `add_filter` to add a filter.

### Current Behavior

```
class FilterChain:
...
    def __init__(self, filters: List[URLFilter] = None):
        self.filters = tuple(filters or [])  # Immutable tuple for speed
        ...

    def add_filter(self, filter_: URLFilter) -> "FilterChain":
        """Add a filter to the chain"""
        self.filters.append(filter_)
        return self  
```
```
E   AttributeError: 'tuple' object has no attribute 'append'
```

### Is this reproducible?

Yes

### Steps to Reproduce

```bash
FilterChain([...]).add_filter([...])
```

nitesh-77:
## Description
Fixes #1753

The `FilterChain` class initializes `self.filters` as a tuple for immutability, but the `add_filter` method was attempting to use `.append()`, which is a list method, causing an `AttributeError`.

## Changes
- Modified `add_filter` to create a new tuple `self.filters + (filter_,)` instead of mutating the existing one.
- This preserves the design intent (immutability) while fixing the crash.

## Verification
- Reproduced the crash with a script.
- Verified that `add_filter` now correctly adds a filter to the chain without raising an error.

### DIFF
--- a/crawl4ai/deep_crawling/filters.py
+++ b/crawl4ai/deep_crawling/filters.py
@@ -85,7 +85,7 @@ class FilterChain:
 
     def add_filter(self, filter_: URLFilter) -> "FilterChain":
         """Add a filter to the chain"""
-        self.filters.append(filter_)
+        self.filters = self.filters + (filter_,)
         return self  # Enable method chaining
 
     async def apply(self, url: str) -> bool:


ntohidi:
thanks for your contribution 💜

Issue #975, opened 2025-04-11
[Bug]: Batch Processing with MemoryAdaptiveDispatcher Missing Crawl Outputs


jacobshenn:
### crawl4ai version

0.5.0

### Expected Behavior

All 680 product URLs passed to crawler.arun_many() should produce a corresponding result.extracted_content if the crawling and extraction process succeeds. Each result should be saved to MongoDB as a new document. I have 680 unique URL's that should output 680 unique MongoDB documents. 

### Current Behavior

I have 680 unique URL's that output 540 MongoDB documents. The MemoryAdaptiveDispatcher is not indicating any errors processing the URL's and all URL's were displayed to be processed. Is this a concurrent processing issue or a race condition issue?

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- A list of 680 product URLs loaded from product_links.json
- This bug happens with any large URL set that I run concurrent processing on.
```

### Steps to Reproduce

```bash

```

### Code snippets

```python
import os
import asyncio
import json
from pydantic import BaseModel
from datetime import datetime, UTC
from crawl4ai import (
    AsyncWebCrawler, BrowserConfig, CrawlerRunConfig, CacheMode,
    LLMConfig, CrawlerMonitor, DisplayMode
)
from crawl4ai.async_dispatcher import MemoryAdaptiveDispatcher
from crawl4ai.extraction_strategy import LLMExtractionStrategy
from pymongo import MongoClient

# MongoDB connection
client = MongoClient("YOUR_MONGO_URI")
collection = client["your_db"]["your_collection"]

class ProductMetadata(BaseModel):
    # Sample Schema

async def main():
    # Load URLs
    with open('product_links.json', 'r') as f:
        product_urls = json.load(f).get('validated_links', [])
    
    browser_cfg = BrowserConfig(
        browser_type="chromium",
        headless=True,
        java_script_enabled=True
    )

    llm_strategy = LLMExtractionStrategy(
        llm_config=LLMConfig(
            provider="openai/gpt-4o",
            api_token="YOUR_API_KEY",
            base_url="https://your-llm-api-url.com"
        ),
        schema=ProductMetadata.model_json_schema(),
        extraction_type="schema",
        instruction="(Sanitized prompt omitted for brevity)",
        chunk_token_threshold=1000,
        apply_chunking=False,
        input_format="markdown",
        extra_args={"temperature": 0.1, "max_tokens": 1000}
    )

[...]

PR #1763, opened 2026-02-17
Fix return in finally block silently suppressing exceptions

Otman404:
## Summary

fixes(https://github.com/unclecode/crawl4ai/issues/975)

Move `return results` out of the `finally` block in `MemoryAdaptiveDispatcher.run_urls` since it was silently suppressing any exception propagating from the `try` or `except` blocks

1. `crawl4ai/async_dispatcher.py` - Moved `return results` from inside the `finally` block to after it in `MemoryAdaptiveDispatcher.run_urls`

### before
```python
        finally:
            # Clean up
            memory_monitor.cancel()
            if self.monitor:
                self.monitor.stop()
            return results
```
### after
```python
        finally:
            # Clean up
            memory_monitor.cancel()
            if self.monitor:
                self.monitor.stop()
        return results
```
2. same file - Re-raise exceptions after logging in the except block

### before
```python
  except Exception as e:
      if self.monitor:
          self.monitor.update_memory_status(f"QUEUE_ERROR: {str(e)}")
```

### after
```python
  except Exception as e:
      if self.monitor:
          self.monitor.update_memory_status(f"QUEUE_ERROR: {str(e)}")
      raise
```

## How Has This Been Tested?

9/10 existing tests pass. The one failure (`test_monitor_integration`)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/crawl4ai/async_dispatcher.py
+++ b/crawl4ai/async_dispatcher.py
@@ -458,7 +458,8 @@ class MemoryAdaptiveDispatcher(BaseDispatcher):
 
         except Exception as e:
             if self.monitor:
-                self.monitor.update_memory_status(f"QUEUE_ERROR: {str(e)}")                
+                self.monitor.update_memory_status(f"QUEUE_ERROR: {str(e)}")
+            raise       
         
         finally:
             # Clean up


unclecode:
Thanks @Otman404 — this has been merged into `develop` and will be included in the next release. Good catch on the return-in-finally bug. We've added you to CONTRIBUTORS.md.

Issue #1731, opened 2026-01-20
[Bug]: Markdown tables missing outer pipe delimiters - causes parser compatibility issues


PatD42:
### crawl4ai version

0.8.0 (also confirmed in 0.7.4)

### Expected Behavior

Markdown tables should include leading and trailing pipe (`|`) delimiters on every row for maximum compatibility with markdown parsers. While this is technically valid GitHub Flavored Markdown (GFM), it causes parsing issues with many markdown processors and violates the recommended best practice for compatibility.

### Current Behavior

  Parameter | Guideline | Common Sources | Health Considerations | Comments
  ---|---|---|---|---
  Arsenic (2006) | 0.010 ALARA | Naturally occurring | Cancer risk | Keep as low as possible

  Issues this causes:
  1. Many markdown parsers fail to render these tables correctly (https://github.com/showdownjs/showdown/issues/230)
  2. Static site generators (Jekyll, Hugo) have inconsistent rendering
  3. Some IDEs and markdown previewers don't recognize the table structure
  4. LLM processing may have reduced accuracy with inconsistent formatting
  5. Potential data loss when parsers misalign columns

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
Any webpage containing HTML `<table>` elements will produce markdown tables with missing outer pipes.

**Test URL used:**
https://www.canada.ca/en/health-canada/services/environmental-workplace-health/reports-publications/water-quality/guidelines-canadian-drinking-water-quality-summary-table.html

This page contains multiple tables with 5-7 columns and 100+ rows. All tables exhibit the issue.
```

### Steps to Reproduce

```bash
1. Install crawl4ai 0.8.0
  2. Crawl any webpage with HTML tables
  3. Examine the generated markdown
  4. Observe that table rows do not start/end with `|`

  **Specific test:**
  - Crawl the Canadian drinking water quality page
  - Check Table 2 (Chemical Parameters) - 97 rows, all missing outer pipes
  - Verify separator row format: `---|---|---|---|---` instead of `|---|---|---|---|---|`
```

### Code snippets

```python
import asyncio
  from crawl4ai import AsyncWebCrawler, CrawlerRunConfig, CacheMode

  async def test_table_format():
      """Test script demonstrating missing outer pipes in markdown tables"""

      url = "https://www.canada.ca/en/health-canada/services/environmental-workplace-health/reports-publications/water-quality/guidelines-canadian-drinking-water-quality-summary-table.html"

[...]

PR #1752, opened 2026-02-08
Fix: Add leading and trailing pipes to markdown tables for GFM compliance

PatD42:
## Summary
Fixes #1731 - Adds leading and trailing pipe delimiters to markdown tables for GitHub Flavored Markdown (GFM) compliance, with proper cell spacing, caption handling, and pad_tables compatibility.

## List of files changed and why
- `crawl4ai/html2text/__init__.py` (lines 680-756) - Fixed table generation logic for GFM compliance
- `tests/test_table_gfm_compliance.py` - 13 comprehensive unit tests

### Before (upstream 0.8.0)
```markdown
Parameter| MAC| AO  
---|---|---  
Aluminum| 2.9 mg/L| 0.1 mg/L  
```

### After
```markdown
| Parameter | MAC | AO |  
| --- | --- | --- |  
| Aluminum | 2.9 mg/L | 0.1 mg/L |  
```

#### 1. Cell pipe delimiters (GFM mode, `pad_tables=False`)
- First cell: `| ` (leading pipe + space)
- Subsequent cells: ` | ` (space + pipe + space) — proper spacing on both sides
- Row end: ` |` (trailing pipe)
- Separator: `| --- | --- |` (with boundary pipes)

#### 2. Table starts on own line
- Added `self.soft_br()` at `<table>` start for non-padded tables
- Tables nested in lists, divs, or after inline content now start on a new line

#### 3. `<caption>` tag handling
- Added `</caption>` handler that outputs a line break
- Prevents caption text from merging inline with the header row
- Real-world example: Canada.ca water quality tables use `<caption>` extensively

#### 4. `pad_tables=True` preserved
- `pad_tables=True` already produces correct GFM output via the TABLE_MARKER post-processor
- Pipe additions are gated on `not self.pad_tables` to avoid double pipes
- Upstream `pad_tables` behavior is completely unchanged

## Why This Fix (Not `pad_tables=True`)
- `pad_tables=False` is the default setting
- `pad_tables` is designed for column alignment, not structural GFM compliance
- Table generation should produce correct output by default
- This fix works correctly with both `pad_tables` settings

### Unit Tests (13 tests covering all scenarios):

**GFM compliance (pad_tables=False):**
- `test_table_has_leading_pipes` - All rows start with `|`
- `test_table_has_trailing_pipes` - All rows end with `|`
- `test_cells_have_space_padding` - Cells have spaces on both sides: `| A | B |` not `| A| B |`
[...]

PR #1764, opened 2026-02-18
Fix: Add leading/trailing pipes to GFM tables (pad_tables=False)

PatD42:
## Summary

When `pad_tables=False` (the default), `html2text` generates table rows without leading and trailing pipe delimiters, producing non-compliant GFM markdown that breaks downstream parsers:

```
Before: A | B | C
After:  | A | B | C |
```

This PR fixes the default table output to produce valid [GitHub Flavored Markdown tables](https://github.github.com/gfm/#tables-extension-).

**Supersedes #1752** (which was auto-closed due to a git history issue).

## Changes

- **Cell delimiters**: Leading `| ` on first cell, ` | ` between subsequent cells
- **Trailing pipe**: ` |` at end of each row
- **Separator row**: `| --- | --- |` format instead of `---|---`
- **Table line start**: `soft_br()` at `<table>` ensures table starts on its own line
- **Caption handling**: `<caption>` text no longer merges inline with the header row
- **pad_tables preserved**: All changes guarded by `not self.pad_tables` — existing `pad_tables=True` behavior is completely unchanged

## Test plan

- [x] 13 unit tests in `tests/test_table_gfm_compliance.py`
  - 10 tests for GFM compliance (leading/trailing pipes, spacing, separator, multirow, single column, empty cells, line start, nested tables, captions)
  - 3 regression tests for `pad_tables=True` (aligned output, no double pipes, separator format)
- [x] Tested against real-world HTML tables (Health Canada water quality guidelines — complex tables with captions, 75+ rows)

Fixes: #1731

### DIFF
--- a/crawl4ai/html2text/__init__.py
+++ b/crawl4ai/html2text/__init__.py
@@ -677,6 +677,11 @@ class HTML2Text(html.parser.HTMLParser):
                     self.o(str(li.num) + ". ")
                 self.start = True
 
+        if tag == "caption":
+            if not start:
+                # Ensure caption text ends on its own line before table rows
+                self.soft_br()
+
         if tag in ["table", "tr", "td", "th"]:
             if self.ignore_tables:
                 if tag == "tr":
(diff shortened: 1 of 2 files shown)


ntohidi:
@PatD42 Thanks for your contribution. Your name will be added to the contributors for our next release (0.8.1) 💜

PR #1768, opened 2026-02-21
Fix cascading context crash under concurrent crawl load

Martichou:
### Problem

When multiple URLs are crawled concurrently and share a BrowserContext (same config signature), all crawls may fail with:
```
BrowserContext.add_init_script: Target page, context or browser has been closed
```
…due to only one URL (e.g. a slow/timing-out site) which was problematic.

Example failure logs — 10 concurrent crawls, all failing within seconds of each other:
```
[ERROR] Error in crawl: BrowserContext.add_init_script: Target page, context or browser has been closed
[ERROR] Error in crawl: BrowserContext.add_init_script: Target page, context or browser has been closed
[ERROR] Error in crawl: BrowserContext.add_init_script: Target page, context or browser has been closed
... (repeated for all concurrent URLs)
```
See detailed logs: [logs.log](https://github.com/user-attachments/files/25454637/logs.log)

### Root cause

`_crawl_web()` in `async_crawler_strategy.py` called `context.add_init_script()` on every crawl for the `override_navigator`, `simulate_user` or `magic` and `shadow dom`. These same scripts are already injected once per context in `BrowserManager.setup_context()`.

The issue is that `add_init_script` is append-only, Playwright has not API to dedup or replace scripts. This causes:

1. Script accumulation: With N concurrent crawls on one context, 2 + 2N init scripts pile up (2 from setup_context + 2 per `_crawl_web` call). On context reuse across batches, this grows unbounded.
2. Memory/CPU pressure: Every new page or frame in the context must execute all accumulated scripts. With 10+ renderer processes each running dozens of duplicate scripts, Chromium hits internal resource limits.
3. Silent context death: Chromium kills the overloaded context without a graceful error. Any concurrent page on that context, including ones that were crawling perfectly fine, immediately gets "Target page, context or browser has been closed".
4. Cascading failure: Because `add_init_script()` is the first context-level operation in `_crawl_web()`, it's the first call to discover the context is dead, producing the distinctive error message across all concurrent tasks.

Removing the add_init_script per scrape fixes the issue has tested using the same script as in https://github.com/unclecode/crawl4ai/pull/1640 (link to direct message https://github.com/unclecode/crawl4ai/pull/1640#issuecomment-3873561529).

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -579,11 +579,19 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             #     [{"name": "cookiesEnabled", "value": "true", "url": url}]
             # )
 
-            # NOTE: navigator_overrider and shadow-DOM init scripts are
-            # already injected once per context in BrowserManager.setup_context().
-            # Do NOT call context.add_init_script() here — it mutates the
-            # shared context from concurrent tasks and causes cascading
-            # "Target page, context or browser has been closed" failures.
+            # Handle navigator overrides (page-scoped to avoid mutating
+            # the shared context from concurrent tasks)
+            if config.override_navigator or config.simulate_user or config.magic:
+                await page.add_init_script(load_js_script("navigator_overrider"))
+
+            # Force-open closed shadow roots when flatten_shadow_dom is enabled
+            if config.flatten_shadow_dom:
+                await page.add_init_script("""
+                    const _origAttachShadow = Element.prototype.attachShadow;
+                    Element.prototype.attachShadow = function(init) {
+                        return _origAttachShadow.call(this, {...init, mode: 'open'});
+                    };
+                """)
 
             # Call hook after page creation
             await self.execute_hook("on_page_context_created", page, context=context, config=config)
--- a/crawl4ai/browser_manager.py
+++ b/crawl4ai/browser_manager.py
@@ -1211,23 +1211,9 @@ class BrowserManager:
                 ]
             )
 
-        # Handle navigator overrides
-        if crawlerRunConfig:
-            if (
-                crawlerRunConfig.override_navigator
-                or crawlerRunConfig.simulate_user
-                or crawlerRunConfig.magic
-            ):
-                await context.add_init_script(load_js_script("navigator_overrider"))
-
-        # Force-open closed shadow roots when flatten_shadow_dom is enabled
-        if crawlerRunConfig and crawlerRunConfig.flatten_shadow_dom:
-            await context.add_init_script("""
-                const _origAttachShadow = Element.prototype.attachShadow;
-                Element.prototype.attachShadow = function(init) {
-                    return _origAttachShadow.call(this, {...init, mode: 'open'});
-                };
-            """)
+        # NOTE: navigator_overrider and shadow-DOM init scripts are now
+        # injected per-page in _crawl_web() via page.add_init_script()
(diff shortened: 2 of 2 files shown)


Martichou:
As per Copilot comments, not sure what you prefer here @unclecode, maybe it's ok to just switch this to `page.add_init_context`. It's up to you, you have better understanding of the inner working here :) 

Edit: pushed an update, let's see :)

unclecode:
@Martichou  Thanks for the excellent bug report and root cause analysis - the diagnosis of unbounded `add_init_script()` accumulation causing cascading context death is spot-on. Really well investigated.

However, I can't merge the fix as-is because switching from `context.add_init_script()` to `page.add_init_script()` introduces a regression for anti-bot scenarios:

**Why `context` scope matters:**
- `context.add_init_script()` runs on **all pages, iframes, and popups** in that context
- `page.add_init_script()` only runs on that specific page and its frames - **not on popups/new tabs**
- Many anti-bot systems (Cloudflare, Akamai) open popup windows for verification. With `page.add_init_script()`, those popups would have a naked `navigator.webdriver = true`, instantly exposing automation

**Why the scripts exist in two places:**
The duplication wasn't accidental. `setup_context()` is called without `crawlerRunConfig` for managed-browser, persistent-context, and CDP paths - so the init scripts don't get injected there. The `_crawl_web()` copy acts as a fallback for those paths. The bug is that for standard/isolated-context paths, both locations fire on the same context.

**The fix I'm going with:**
Flag-based dedup - after injecting a script, we set `context._crawl4ai_nav_overrider_injected = True` (and same for shadow DOM). Before injecting, we check the flag. This way:
- Scripts are injected **at most once** per context regardless of path
- Context-level scope is preserved (popups stay covered)
- No accumulation under concurrent load

Will be in the next release. Thanks again for catching this - the concurrency failure mode you documented will help others who hit this before the update.

Martichou:
@unclecode this is actually quite a better changes that you proposed indeed! :) I do think that currently this was the last annoying bug I had on my list regarding Crawl4AI with a high-concurrency environment. I've attached some images which show the before/after which confirms the drastic improvement!

I hope you'll also enjoy all these improvements in your new Cloud offering for C4AI 🙏
Gonna close this issue as it's fixed with your changes, thanks for your time and building this up 😄 Really enjoyed the collaboration, hope this will happens again.
If you ever want to chat or need help on something, hit me up (I have all the info you may need on my Github).



Before (regularly hitting "high" avg duration & hard timeout of 2 minutes):
Screenshot 2026-02-24 at 11 58 41

After (including the recycling fixes):
Screenshot 2026-02-24 at 11 58 57

unclecode:
@Martichou Those before/after charts speak for themselves - from hitting 2-minute timeouts down to a steady ~8s avg. That's a massive difference in a production high-concurrency setup, and it validates the fix perfectly. Thanks for taking the time to measure and share this.

Really appreciated the quality of your contributions here - the root cause analysis on the `add_init_script` accumulation was thorough and well-documented, and the back-and-forth on the right fix (context-scope vs page-scope) led to a better solution than either of us would've landed on alone. That's exactly how open source should work.

On that note - we're building out the Crawl4AI Cloud service and are starting paid collaborations with engineers who understand the system deeply, especially around concurrency, browser management, and production-scale crawling. Given your experience running C4AI at scale, I think you'd be a great fit.

If you're interested, shoot an email to aravind@crawl4ai.com (cc: unclecode@crawl4ai.com) and we can talk details.

Looking forward to more of this kind of collaboration either way.

PR #1771, opened 2026-02-23
Claude/check fork sync s9 s sz

### LINKED ISSUE #1769: [Bug]:  mcp_bridge: httpx default 5s timeout causes isError on slow LLM-backed endpoints
### crawl4ai version

0.8.0

### Expected Behavior

 The MCP tool call should return the processed markdown content successfully, regardless of how long the LLM-backed endpoint takes to respond. The
   MCP bridge should wait for the server to finish and relay the result back to the client.

### Current Behavior

  When an LLM-backed endpoint (e.g. GET /md?filter=llm) takes longer than 5 seconds to respond, the MCP bridge raises an httpx.TimeoutException due
   to the default 5-second timeout on httpx.AsyncClient(). This exception is not caught by the existing except httpx.HTTPStatusError handler and
  propagates to the MCP framework, which returns isError: true with an empty text body — even though the crawl4ai server completed the request
  successfully.

  The failure is intermittent: it only occurs when the LLM response time exceeds ~5 seconds, making it appear as a random/flaky error.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- MCP tool: md
  - Parameter filter=llm (any LLM-backed content filter)
  - Any URL where the LLM processing time exceeds ~5 seconds (depends on model speed, page length, and server load)
  - Example: a moderately long web page crawled with LLMContentFilter using a slower or remote LLM model

  The bug is not tied to a specific URL — it reproduces on any page where the LLM takes longer than the hardcoded 5-second httpx default timeout.
```

### Steps to Reproduce

```bash
1. Start the crawl4ai Docker server with an LLM provider configured (e.g. OpenAI, Anthropic).
  2. Connect an MCP client (e.g. Claude Desktop) to the crawl4ai MCP bridge.
  3. Call the md MCP tool with filter=llm on any moderately long web page:
  url: https://example.com/some-long-article
  filter: llm
  4. Observe that the MCP tool returns isError: true with empty content.
  5. Check the crawl4ai server logs — the request completed successfully (e.g. fit_md_len=2076), confirming the server finished but the bridge
  timed out before receiving the response.

  Note: The issue is intermittent. It reproduces reliably when the LLM response takes longer than ~5 seconds. Using a slower model or a longer page
   increases the chance of triggering it.
```

### Code snippets

```python
# BEFORE (buggy)
  # httpx.AsyncClient() uses a default 5-second timeout.
  # When the LLM takes longer, httpx raises TimeoutException.
  # This is NOT a subclass of HTTPStatusError, so it is never caught —
  # it propagates to the MCP framework as isError: true with empty text.
  async with httpx.AsyncClient() as client:
      try:
          r = await client.get(url, params=kwargs)
          ...
      except httpx.HTTPStatusError:
          # TimeoutException does NOT match here — it bypasses this handler
          ...

  # AFTER (fixed)
  # timeout=None disables the client-side timeout entirely.
  # The crawl4ai server manages its own internal timeouts.
  async with httpx.AsyncClient(timeout=None) as client:
      try:
          r = await client.get(url, params=kwargs)
          ...
      except httpx.HTTPStatusError:
          ...
```

### OS

Linux

### Python version

3.13

### Browser

_No response_

### Browser version

_No response_

_No response_

hafezparast:
**Summary:**

Fixes #1769

The MCP bridge's `httpx.AsyncClient()` uses a default 5-second timeout. When LLM-backed endpoints (e.g. `md` tool with `filter=llm`) take longer to respond, `httpx.TimeoutException` is raised. Since this isn't a subclass of `HTTPStatusError`, it bypasses the existing exception handler and propagates to the MCP framework as an unhandled error — even though the server successfully completed the request.

This PR adds a configurable `timeout` parameter (defaulting to `None` / no limit) and catches `TimeoutException` explicitly, surfacing it as a proper HTTP 504 instead of an opaque crash.

**List of files changed and why:**

- `deploy/docker/mcp_bridge.py` — Added `timeout` parameter to `_make_http_proxy()` and `attach_mcp()`, passed it to `httpx.AsyncClient(timeout=...)`, and added an `except httpx.TimeoutException` handler that raises `HTTPException(504)`.

**How Has This Been Tested?**

- Reviewed all call sites of `attach_mcp()` (only `server.py:829`) — the default `timeout=None` is backward-compatible, so existing behavior is preserved without changes to `server.py`.
- Verified that `httpx.AsyncClient(timeout=None)` disables all timeouts, allowing the crawl4ai server to manage its own timeout logic.
- Verified that passing a numeric value (e.g. `timeout=300`) would enforce a 5-minute upper bound if desired.

**Checklist:**

---

### DIFF
--- a/deploy/docker/mcp_bridge.py
+++ b/deploy/docker/mcp_bridge.py
@@ -37,7 +37,7 @@ def mcp_tool(name: str | None = None):
     return deco
 
 # ── HTTP‑proxy helper for FastAPI endpoints ─────────────────────
-def _make_http_proxy(base_url: str, route):
+def _make_http_proxy(base_url: str, route, *, timeout: float | None = None):
     method = list(route.methods - {"HEAD", "OPTIONS"})[0]
     async def proxy(**kwargs):
         # replace `/items/{id}` style params first
@@ -49,7 +49,7 @@ def _make_http_proxy(base_url: str, route):
                 kwargs.pop(k)
         url = base_url.rstrip("/") + path
 
-        async with httpx.AsyncClient() as client:
+        async with httpx.AsyncClient(timeout=timeout) as client:
             try:
                 r = (
                     await client.get(url, params=kwargs)
@@ -61,6 +61,8 @@ def _make_http_proxy(base_url: str, route):
             except httpx.HTTPStatusError as e:
                 # surface FastAPI error details instead of plain 500
                 raise HTTPException(e.response.status_code, e.response.text)
+            except httpx.TimeoutException:
+                raise HTTPException(504, "upstream request timed out")
     return proxy
 
 # ── main entry point ────────────────────────────────────────────
@@ -70,6 +72,7 @@ def attach_mcp(
     base: str = "/mcp",
     name: str | None = None,
     base_url: str,              # eg. "http://127.0.0.1:8020"
+    timeout: float | None = None,  # httpx timeout in seconds; None = no limit
 ) -> None:
     """Call once after all routes are declared to expose WS+SSE MCP endpoints."""
     server_name = name or app.title or "FastAPI-MCP"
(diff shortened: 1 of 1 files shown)


ntohidi:
Thanks for your contribution

PR #1772, opened 2026-02-24
fix: resolve MCP bridge timeout on slow LLM endpoints (#1769)

zhu6201976:
Problem:
The current mcp_bridge uses httpx.AsyncClient with its default 5-second timeout. When interfacing with LLM-backed endpoints (especially complex reasoning tasks or slow TTFT), this frequently triggers a TimeoutException, resulting in an isError response even when the backend is functioning correctly.

Solution:

Increased the httpx timeout to 300 seconds (5 minutes) to accommodate typical LLM processing times.

Implemented a structured httpx.Timeout(300.0, connect=10.0) for better connection management.

Added explicit error handling for httpx.TimeoutException to provide clearer feedback to the user.

Impact:
Improved stability for MCP-based workflows, particularly when using agents or slow upstream LLM APIs.

How to test:
Set up an MCP bridge pointing to a mock endpoint with a >5s delay.

Verified that the request now completes successfully instead of timing out at 5s.

Verified that requests exceeding 300s return the new descriptive error message.

### DIFF
--- a/deploy/docker/mcp_bridge.py
+++ b/deploy/docker/mcp_bridge.py
@@ -36,6 +36,11 @@ def mcp_tool(name: str | None = None):
         return fn
     return deco
 
+# Timeouts for LLM/backend calls (Issue #1769: avoid default 5s causing LLM errors)
+LLM_REQUEST_TIMEOUT = 300.0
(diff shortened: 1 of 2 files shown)


unclecode:

Issue #1754, opened 2026-02-10
[Bug]: Crawl4Ai pod goes unresponsive after few days of successful crawl


Jetty53:
### crawl4ai version

v0.8.0

### Expected Behavior

Craw4Ai should not deteriorate after some days of crawling.

### Current Behavior

After a few days of successfull crawling, I am getting this log in the crawl4ai pod

2026-02-10 06:59:17,220 - server - WARNING - Timeline update timeout after 4s
2026-02-10 06:59:26,221 - server - WARNING - Timeline update timeout after 4s
2026-02-10 06:59:35,223 - server - WARNING - Timeline update timeout after 4s
2026-02-10 06:59:44,225 - server - WARNING - Timeline update timeout after 4s
2026-02-10 06:59:53,226 - server - WARNING - Timeline update timeout after 4s
2026-02-10 07:00:02,229 - server - WARNING - Timeline update timeout after 4s
2026-02-10 07:00:11,231 - server - WARNING - Timeline update timeout after 4s

and this is forever. Crawl4ai pod becomes unresponsive. Can't start a new crawl, the ongoing crawl can't move further.

Only solution at this point is either restart the pod, or flush-all the Crawl4ai Redis. But thats not a proper solution. 

### Is this reproducible?

No

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python

```

### OS

Linux

### Python version

3.12

ntohidi:
Discord messages: https://discord.com/channels/1278297938551902308/1368102652910632980/1470677153438892094

Ahmed-Tawfik94:
## Root Cause Analysis

### Primary cause:

**Deadlock in asyncio.Lock during timeline update**

The issue is in `deploy/docker/monitor.py`, line 154 in the `update_timeline()` method:

```python
async def update_timeline(self):
    """Update timeline data points (called every 5s)."""
    now = time.time()
    mem_pct = get_container_memory_percent()

    # Clean old entries (keep last 5 minutes)
    self._cleanup_old_entries(max_age_seconds=300)

    # Count requests in last 5s
    recent_reqs = sum(1 for req in self.completed_requests
                     if now - req.get("end_time", 0) < 5)

    # Browser counts (acquire lock to prevent race conditions)
    from crawler_pool import PERMANENT, HOT_POOL, COLD_POOL, LOCK
    async with LOCK:  # <-- POTENTIAL DEADLOCK HERE
[...]

PR #1773, opened 2026-02-24
perf: enhance monitor stability with non-blocking LOCK acquisition (#1754)

zhu6201976:
Summary:
While the recent changes in develop introduce a global wait_for timeout for update_timeline, the monitor still attempts to "queue" for the global LOCK. Under extreme load, this can lead to unnecessary resource contention.

Improvements:

Non-blocking Stats: Changed LOCK acquisition within monitor.py to a fail-fast mechanism (0.5s timeout). If the crawler pool is busy, the monitor skips the current data point instead of adding to the lock queue.

System Resilience: Ensures that telemetry collection never interferes with the core mission-critical path (browser lifecycle management).

Error Silencing: Downgraded lock-timeout warnings to debug logs to avoid cluttering production logs during expected high-load spikes.

Why this matters:
In high-concurrency enterprise environments, skipping a single monitoring heartbeat is preferable to adding latency to the asyncio event loop.

### DIFF
--- a/deploy/docker/crawler_pool.py
+++ b/deploy/docker/crawler_pool.py
@@ -32,9 +32,12 @@ def _is_default_config(sig: str) -> bool:
     return sig == DEFAULT_CONFIG_SIG
 
 async def get_crawler(cfg: BrowserConfig) -> AsyncWebCrawler:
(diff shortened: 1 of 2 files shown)


unclecode:

PR #1788, opened 2026-03-01
fix: guard against None LLM content and propagate finish_reason

### LINKED ISSUE #1606: [Bug]: The maxtokens setting is too small and there is no correct prompt
### crawl4ai version

0.7.6

### Expected Behavior

 using the LLMExtractionStrategy,  model choice gemini

### Current Behavior

When using the LLMExtractionStrategy, if the maxtokens are set too small, the context window is insufficient to carry the output content of the LLM, so there is no return value. However, the returned Json of the LLM, such as "finish_reason":"MAX_TOKENS", does not return to the location of the outermost method call, making it impossible for the caller to determine the core of the problem

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python
result = await crawler.arun(
            url="example.com", config=crawler_config
        )

print(result.extracted_content)
//output
[
    {
        "index": 0,
        "error": true,
        "tags": [
            "error"
        ],
        "content": "'NoneType' object has no attribute 'startswith'"
    }
]
```

### OS

Windows

### Python version

3.12.9

### Browser

_No response_

### Browser version

_No response_

[
    {
        "index": 0,
        "error": true,
        "tags": [
            "error"
        ],
        "content": "'NoneType' object has no attribute 'startswith'"
    }
]

Br1an67:
## Summary
When `max_tokens` is set too small, the LLM may return `None` content with `finish_reason=MAX_TOKENS`. This causes a crash: `'NoneType' object has no attribute 'startswith'`.

This PR:
1. Guards against `None` content from LLM responses — returns an error block with the `finish_reason` so callers can diagnose the issue
2. Protects the fallback `split_and_parse_json_objects` path from `None` content

Fixes #1606

- `crawl4ai/extraction_strategy.py` — Added None content guard with finish_reason propagation in both sync and async extraction paths

## How Has This Been Tested?
Verified that when content is None, a meaningful error block is returned instead of a crash. The error block includes finish_reason for diagnostic purposes.

## Checklist:

### DIFF
--- a/crawl4ai/extraction_strategy.py
+++ b/crawl4ai/extraction_strategy.py
@@ -676,7 +676,11 @@ class LLMExtractionStrategy(ExtractionStrategy):
                 content = response.choices[0].message.content
                 blocks = None
 
-                if self.force_json_response:
+                if not content:
+                    finish_reason = getattr(response.choices[0], "finish_reason", "unknown")
+                    blocks = [{"index": 0, "error": True, "tags": ["error"],
+                               "content": f"LLM returned no content (finish_reason: {finish_reason})"}]
+                elif self.force_json_response:
                     blocks = json.loads(content)
                     if isinstance(blocks, dict):
                         # If it has only one key which calue is list then assign that to blocks, exampled: {"news": [..]}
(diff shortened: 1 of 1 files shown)


unclecode:
Thanks @Br1an67 — merged into `develop`, will be in the next release. The None content guard prevents crashes when LLMs hit content filters or token limits. You've been added to CONTRIBUTORS.md.

PR #1795, opened 2026-03-01
fix: require api_token for /token endpoint when configured

### LINKED ISSUE #1627: [Bug]: endpoint /token does not require credentials
### crawl4ai version

0.7.6

### Expected Behavior

When running crawl4ai from the docker image with a security config of 
`security:
  enabled: true
  jwt_enabled: true
` I would expect it is required to present a secret to the endpoint POST /token in order to obtain a temporary access token.

### Current Behavior

The endpoint POST /token can be called without any credentials and it still returns a valid token that can subsequently be used to call protected endpoints such as POST /crawl. This renders the security useless in my understanding, because anyone with network access to the API alone can use it.

In previous versions we had CRAWL4AI_API_TOKEN to protect the API from unauthorised usage. But sadly, this simple but effective mechanism has been removed from later versions.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
Build the Docker image using a config.yml with 
security:
  enabled: true
  jwt_enabled: true

Run the Docker image

Go to http://127.0.0.1:11235/docs#/default/get_token_token_post

Obtain a token

Use the token to call POST /crawl
```

### Code snippets

```python
deploy/docker/server.py at https://github.com/unclecode/crawl4ai/blob/main/deploy/docker/server.py#L288 does not require any credentials to obtain a valid JWT token for subsequent access to protected endpoints:

@app.post("/token")
async def get_token(req: TokenRequest):
    if not verify_email_domain(req.email):
        raise HTTPException(400, "Invalid email domain")
    token = create_access_token({"sub": req.email})
    return {"email": req.email, "access_token": token, "token_type": "bearer"}
```

### OS

macOS

### Python version

3.10

### Browser

Firefox

### Browser version

_No response_

_No response_

Br1an67:
## Summary

Fixes #1627

The `/token` endpoint issues JWT tokens to anyone with a valid email domain, with no credential check. This adds an `api_token` field to the security config. When set, the `/token` endpoint requires the caller to provide the matching `api_token` in the request body before issuing a JWT.

Existing deployments without `api_token` configured are unaffected (backward compatible).

- `deploy/docker/auth.py` — Added optional `api_token` field to `TokenRequest`
- `deploy/docker/server.py` — Validate `api_token` against config before issuing JWT
- `deploy/docker/config.yml` — Added `api_token` field under `security` section

## How Has This Been Tested?

Verified logic: when `security.api_token` is empty (default), the endpoint works as before. When set, requests without a matching `api_token` get a 401 response.

## Checklist:

### DIFF
--- a/deploy/docker/auth.py
+++ b/deploy/docker/auth.py
@@ -71,3 +71,4 @@ def get_token_dependency(config: Dict):
 
 class TokenRequest(BaseModel):
     email: EmailStr
+    api_token: Optional[str] = None
--- a/deploy/docker/config.yml
+++ b/deploy/docker/config.yml
@@ -44,6 +44,7 @@ rate_limiting:
 security:
   enabled: false
   jwt_enabled: false
+  api_token: ""  # When set, /token endpoint requires this secret to issue JWTs
   https_redirect: false
   trusted_hosts: ["*"]
   headers:
--- a/deploy/docker/server.py
+++ b/deploy/docker/server.py
@@ -303,6 +303,9 @@ logger = logging.getLogger(__name__)
 # ──────────────────────── Endpoints ──────────────────────────
 @app.post("/token")
 async def get_token(req: TokenRequest):
+    expected_token = config.get("security", {}).get("api_token", "")
+    if expected_token and req.api_token != expected_token:
+        raise HTTPException(401, "Invalid or missing api_token")
     if not verify_email_domain(req.email):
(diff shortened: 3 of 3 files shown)


unclecode:
Thanks @Br1an67 - this is a solid security hardening for the Docker deployment. We've merged it into develop and it will be in the next release. We'll acknowledge your contribution.

PR #1796, opened 2026-03-01
fix: use DOMParser instead of innerHTML in process_iframes

### LINKED ISSUE #1582: [Bug]: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML'
### crawl4ai version

0.7.4

### Expected Behavior

As you would expect, I'm building a little python project to crawl and deep crawl websites. For testing purposes I use https://about.google with a depth of 2. 

I expect the crawler to crawl all the page's content without errors when using chromium.

### Current Behavior

Using https://about.google with a depth of 2 several links popping up return the same error several times for different sub pages. This only happens when using chromium. Using firefox it works without any issues.

When comparing the crawled content of both the only difference is, that with chromium embedded youtube videos seem to be missing.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- URL(s): https://about.google/intl/ALL_kr/around-the-globe/local-info
- Settings used:
  - browser config
    - browser_type: "chromium"
    - headless: true
    - user_agent: "Chrome/116.0.0.0"
    - user_agent_mode: "default"
    - user_agent_generator_config:
      - platforms: 'mobile'
      - os: 'Android'
    - viewport_width: 1080
    - viewport_height: 64
    - accept_downloads: false
    - ignore_https_errors: true
    - java_script_enabled: true
    - enable_stealth: true
    - text_mode: false
    - verbose: true
  - run config
    - deep_crawl_strategy
      - BFSDeepCrawlStrategy
      - max_depth: 2,
      - include_external: false
    - only_text: false
    - scan_full_page: true
    - process_iframes: true
    - remove_overlay_elements: false
    - simulate_user: true
    - override_navigator: true
    - score_links: true
    - magic: true
    - cache_mode: 'disabled'
```

### Steps to Reproduce

```bash

```

### Code snippets

```python
async with AsyncWebCrawler(config=self._browser_cfg) as crawler:
    async for result in await crawler.arun(
        url=url,
        config=run_cfg,
    ):
        print(result.url)
```

### OS

Linux 6.6.80-1-MANJARO x86_64

### Python version

3.13.2

### Browser

chromium

### Browser version

chromium-1194

### Additional information

playwright version: 1.55.0

### Error log

> [ERROR]... × Error processing iframe 0: Page.evaluate: TypeError: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML' assignment and no 'default' policy 
for 'TrustedHTML' has been defined.
    at eval (eval at evaluate (:291:30), <anonymous>:4:43)
    at UtilityScript.evaluate (<anonymous>:298:18)
    at UtilityScript.<anonymous> (<anonymous>:1:44)

Br1an67:
## Summary

Fixes #1582

Sites with Trusted Types Content Security Policy (like Google properties) block direct `innerHTML` assignment, causing `TypeError: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML'`. 

Replaced `div.innerHTML = ...` with `DOMParser.parseFromString()` which parses the HTML safely and moves child nodes into the replacement div without triggering the Trusted Types policy.

`crawl4ai/async_crawler_strategy.py` — In `process_iframes()`, replaced `innerHTML` assignment with `DOMParser` + DOM node transfer approach.

## How Has This Been Tested?

Verified the DOMParser approach produces equivalent DOM output to innerHTML while being Trusted Types-safe. DOMParser is supported in all modern browsers.

## Checklist:

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -380,7 +380,11 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
                         () => {{
                             const iframe = document.getElementById('iframe-{i}');
                             const div = document.createElement('div');
-                            div.innerHTML = `{_iframe}`;
+                            const parser = new DOMParser();
+                            const doc = parser.parseFromString(`{_iframe}`, 'text/html');
+                            while (doc.body.firstChild) {{
+                                div.appendChild(doc.body.firstChild);
+                            }}
                             div.className = '{class_name}';
                             iframe.replaceWith(div);
                         }}


unclecode:
Thanks @Br1an67 - clean security fix. DOMParser is the right approach for safely injecting iframe content. We've merged this into develop and it will be in the next release. We'll acknowledge your contribution.

Issue #1087, opened 2025-05-08
[Bug]: cannot crawl all content while scan_full_page=True


duanyr:
### crawl4ai version

0.5.0.post8

### Expected Behavior

get all note-item content when scroll to the last page

### Current Behavior

only obtain the content of the note-item class on the last page

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
URL：https://www.xiaohongshu.com/search_result/?keyword=%25E6%2597%25A0%25E9%2594%25A1%25E6%25B1%2582%25E7%25A7%259F&source=unknown&type=50

Hello,  When I scroll through the list page of the website I want to crawl, the number of note-item in the feeds-container area remains fixed. For example, during the scrolling process, although the data-index in the screenshot keeps increasing, the number of note-item in the feeds-container is fixed. In such a situation, how should I crawl the content of all note-item classes during the scrolling process?
When I execute my code, it will automatically scroll to the last page, but in the end, it will only obtain the content of the note-item class on the last page. Since the content of the previous pages is no longer in the feeds-container, I can't get it.
```

### Steps to Reproduce

```bash

```

### Code snippets

```python
async def extract_rent_user():
    browser_config = BrowserConfig(
        # browser_type="chromium",
        headless=False,# Set to False if you want to see the browser window
        use_managed_browser=True,  # Required for persistent profiles
        user_data_dir="/root/.crawl4ai/profiles/profile_20250415_120515_1a63da",
    )
    crawler_config = CrawlerRunConfig(
        cache_mode=CacheMode.BYPASS,
        wait_for_images=True,
        scan_full_page=True,  # Tells the crawler to try scrolling the entire page
        scroll_delay=0.5,  # Delay (seconds) between scroll steps
        # verbose=True,
        js_code="window.scrollTo(0, document.body.scrollHeight);",
        extraction_strategy=JsonCssExtractionStrategy(
            schema={
                "name": "XHS Rent User Search Results",
                "baseSelector": ".feeds-container .note-item",
                "fields": [
                    {
                        "name": "user_name",
                        "selector": ".footer .author .name",
                        "type": "text"
                    },
                    {
                        "name": "card_time",
                        "selector": ".footer .author .time",
                        "type": "text"
                    },
                    {
                        "name": "title",
                        "selector": ".title span",
                        "type": "text"
                    },
                    {
                        "name": "room_detail_url",
                        "selector": ".cover.mask.ld",
                        "type": "attribute",
[...]

PR #1853, opened 2026-03-23
fix: scan_full_page captures all content on virtual-scroll pages (#731)

### LINKED ISSUE #731: [Bug]: When extracting data with scroll_full_page, only the final elements get parsed
### crawl4ai version

0.4.248

### Expected Behavior

I'm crawiling Twitter, specifically the "following" section of a profile. I have a css selector for the relevant data (user's names and bios) and set up a JsonCssExtractionStrategy. If I don't use scroll_full_page, I understandably only expect to get the first N user profiles. But if I do enable scroll_full_page, I expect the returned data to contain the list to the fullest extent as visible when browsing manually.

### Current Behavior

When not using scroll_full_page, I do get the first 16 profiles in this case. However, when setting scroll_full_page, I only get the LAST 12. It's important to note that there are over 40 profiles listed, so none of the 12 profiles intersect with the first 16. I checked the result's html property and it does only contain information about the last 12. However, strangely the screenshot saved contains all the profiles.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
https://x.com/SomeTwitterProfile/following
```

### Steps to Reproduce

```bash
Execute the following snippet with both scoll_full_page on and off.
```

### Code snippets

```python
site_url = "https://x.com/elonmusk/following"

import asyncio
from crawl4ai import AsyncWebCrawler, CacheMode, BrowserConfig, CrawlerRunConfig
from crawl4ai.markdown_generation_strategy import DefaultMarkdownGenerator
from crawl4ai.extraction_strategy import JsonCssExtractionStrategy
import json


async def main():
    schema = {
        "name": "Followers",
        "baseSelector": "button[data-testid='UserCell']",
        "fields": [
            {
              "name": "name",
              "selector": "span",
              "type": "text"
	    },
            {
              "name": "handle",
              "selector": 'a[role="link"] > div > div[dir="ltr"]:only-child > span ',
              "type": "text"
	    },
            {
              "name": "bio",
              "selector": "div[dir='auto'] > span",
              "type": "html",
	     }
                        
	 ]
    }
    
    extraction_strategy = JsonCssExtractionStrategy(schema, verbose=True)
    
    browser_conf = BrowserConfig(
        extra_args=['--disable-web-security'],
        cookies=[
            {"name": "auth_token", "value": "YOURAUTHTOKEN", "domain": ".x.com", "path": "/"},
        ],
    )
    
        cache_mode=CacheMode.BYPASS, #        cache_mode=CacheMode.BYPASS/DISABLED,
        screenshot=True,
        wait_for="css:button[data-testid='UserCell']:nth-child(1)",
        page_timeout=5000,
        extraction_strategy=extraction_strategy,
        # scan_full_page=True,
        scroll_delay=1.5,
    )

    try:
        async with AsyncWebCrawler(
            config=browser_conf,
            verbose=True,
        ) as crawler:
            result = await crawler.arun(
                url=site_url ,
                config=crawler_config,
            )
                
            from base64 import b64decode
            with open("screenshot.png", "wb") as f:
              f.write(b64decode(result.screenshot))
              
            with open("index.html", "w") as f:
              f.write(result.html)
                                
            data = json.loads(result.extracted_content)
            print(f"Extracted {len(data)} users")
            print(json.dumps(data, indent=2) if data else "No data found")
                    
    except Exception as e:
        print(f"Something happened: {e}")
        raise e

if __name__ == "__main__":
    asyncio.run(main())
```

### OS

Linux

### Python version

3.12.9

### Browser

Default

### Browser version

_No response_

_No response_

hafezparast:
## Summary

- **`_handle_full_page_scan`**: installs a `MutationObserver` before scrolling to capture DOM elements that are removed during scroll (virtual scroll recycling). After scrolling, deduplicates against still-visible elements and re-injects accumulated content into a hidden `<div>` so `page.content()` includes everything.
- **`_handle_virtual_scroll`**: falls back to `window.scrollBy()` when `container.scrollTop` has no effect (window-level scrolling, e.g. Twitter/X). Updates end-of-scroll detection accordingly.
- Non-virtual-scroll pages are unaffected: if no elements are removed during scrolling, no injection occurs.

**Before fix:** `scan_full_page=True` on a 50-item virtual-scroll page → only 10 items (last batch)  
**After fix:** all 50 items captured

Fixes #731, related to #1087

## Test plan

- [x] Reproduction test (`tests/test_repro_731.py`) — local virtual-scroll page with 50 items, verifies all 50 captured
- [x] 13 adversarial tests covering: virtual scroll (50/100 items), static pages (no false positives), empty pages, append-based infinite scroll, nested virtual scroll with header/footer preservation, low max_steps, observer cleanup, element ordering, hidden container verification, deduplication
- [x] Full regression suite: 303 passed, 0 regressions (1 pre-existing HuggingFace failure unrelated)

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -1234,53 +1234,181 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             viewport_height = viewport_size.get(
                 "height", self.browser_config.viewport_height
             )
+
+            # Snapshot visible elements before first scroll so we capture
+            # the initial viewport on virtual-scroll pages.
+            await page.evaluate("""() => {
+                window.__c4ai_snapshot = new Map();
+                window.__c4ai_snapshot_fn = (function() {
+                    const MAX_ITEMS = 10000;
+                    function fingerprint(el) {
+                        // Prefer id attribute for uniqueness
+                        if (el.id) return 'id:' + el.id;
+                        // Look for data-* ID attributes (data-id, data-item-id,
+                        // data-user-id, etc.) but skip type/role markers like
+                        // data-testid which are shared across items.
+                        for (const attr of el.attributes) {
+                            if (attr.name.startsWith('data-')
+                                && (attr.name === 'data-id'
+                                    || attr.name.endsWith('-id')
+                                    || attr.name.endsWith('-key'))
+                                && attr.value) {
+                                return 'data:' + attr.name + '=' + attr.value;
+                            }
+                        }
+                        // Fallback: two independent 32-bit hashes combined into
+                        // a 64-bit-equivalent string to avoid collisions at scale.
+                        const tag = el.tagName;
+                        const href = el.getAttribute('href') || '';
+                        const text = (el.textContent || '').trim();
+                        let h1 = 0, h2 = 0x9e3779b9;
+                        for (let i = 0; i < text.length; i++) {
+                            const c = text.charCodeAt(i);
+                            h1 = ((h1 << 5) - h1 + c) | 0;
+                            h2 = ((h2 << 7) ^ (h2 >>> 3) ^ c) | 0;
+                        }
+                        return 'hash:' + tag + ':' + href + ':' + h1 + ':' + h2;
+                    }
+                    // Check if element carries a data-* unique identifier
+                    // (data-item-id, data-user-id, etc.). We do NOT check
+                    // el.id here because structural containers commonly have
+                    // ids (id="feed", id="main") without being content items.
+                    function hasDataId(el) {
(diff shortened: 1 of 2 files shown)


hafezparast:
Reworked based on your review. Force-pushed with the scroll-and-snapshot approach:

- Replaced MutationObserver with a `Map<fingerprint, outerHTML>` loop
- Fingerprinting uses `id`/`data-*` attributes first, content hash fallback
- Accumulator capped at 10K elements
- Kept the `window.scrollBy` fallback
- Real pytest tests with `@pytest.mark.asyncio` — 9 adversarial cases (1000 items, similar text, no data-id, append-only, memory stress)

Sorry about the detour through #1877 — consolidated everything back here.

PR #1877, opened 2026-03-28
fix: scan_full_page captures all content on virtual-scroll pages (#731) v2

### LINKED ISSUE #731: [Bug]: When extracting data with scroll_full_page, only the final elements get parsed

hafezparast:
## Summary

Rework of #1853 based on @unclecode's review. Replaces MutationObserver with a scroll-and-snapshot approach using a `Map` for deduplication.

- **Scroll-and-snapshot**: At each scroll step, walk the live DOM and fingerprint visible elements into a `Map`. No mutation callbacks, no detached-node issues.
- **Smart fingerprinting**: Prioritize `data-id` / `data-*-id` / `data-*-key` attributes (skip `data-testid` which is a shared type marker). Fallback uses dual 32-bit hashes (64-bit equivalent) to avoid collisions at scale.
- **Container detection**: `isItemContainer()` identifies feed/list parents by checking if most children share the same tag. Elements with `data-*-id` are treated as content items and never drilled into.
- **10K element cap**: Prevents unbounded memory growth.
- **`window.scrollBy` fallback**: Both `_handle_full_page_scan` and `_handle_virtual_scroll` fall back to `window.scrollBy` when container scroll doesn't move (e.g., Twitter scrolls the window, not a container).
- **12 adversarial pytest tests** covering: basic recycling, 200/1000-item scale, similar-text collision, no-data-id fallback, append-only scroll, DOM noise, static page regression, scrollBy fallback, rich cards with 4+ children, 20-level deep nesting, and 500-item hash collision resistance.

## What changed vs. v1 (#1853)

| Issue (from review) | v1 (MutationObserver) | v2 (this PR) |
|---|---|---|
| CPU overhead | Observer fires 20-40x per scroll step | Snapshot runs once per scroll step |
| Memory | Unbounded `__c4ai_removed` array | `Map` deduplicates inline, capped at 10K |
| DOM noise | Captures ALL removed nodes (tooltips, ads) | Only reads container children |
| Detached nodes | `innerText` can return empty | Reads from live DOM |
| Fingerprint collision | `substring(0, 200)` collapses similar items | `data-*-id` first, dual-hash fallback |
| `data-testid` bug | Not applicable (different approach) | Explicitly excluded from ID matching |
| Tests | Standalone script, happy path only | 12 pytest tests with adversarial cases |

## Known limitations

These are architectural boundaries of the scroll-and-snapshot approach, not bugs:

1. **Non-window scroll containers**: If a page scrolls inside a specific `<div>` (e.g., Gmail's email list), our window-level scroll won't move it. Use `virtual_scroll_config` with `container_selector` for these cases.
2. **Identical-text items without `data-*-id`**: Items with truly identical `textContent` and no distinguishing attributes hash to the same fingerprint. Only one copy is kept.
3. **Async lazy-loading**: If a page fetches content from an API >200ms after scroll settles, the snapshot may capture the loading state. Increase `scroll_delay` as a workaround.
4. **Shadow DOM**: `el.children` doesn't cross shadow boundaries. Items inside shadow roots are invisible to the walk.
5. **Mixed-tag feed containers**: `isItemContainer` requires >=50% of children to share a tag. Feeds that alternate items with separators or ad sections may not be detected as containers.
6. **Stale `id`-based captures**: Elements with an `id` attribute are fingerprinted as `id:<value>` and captured once. If their content changes during scrolling, the stale version is kept.
7. **Hidden div side effects**: The injected `display:none` div inflates element counts for downstream queries like `querySelectorAll`.

## Test plan

- [x] 12/12 adversarial tests passing locally
- [ ] Verify no regression on existing test suite
- [ ] Test on a real virtual-scroll page (Twitter followers, etc.)

Closes #731

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -1234,53 +1234,181 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             viewport_height = viewport_size.get(
                 "height", self.browser_config.viewport_height
             )
+
+            # Snapshot visible elements before first scroll so we capture
+            # the initial viewport on virtual-scroll pages.
+            await page.evaluate("""() => {
+                window.__c4ai_snapshot = new Map();
+                window.__c4ai_snapshot_fn = (function() {
+                    const MAX_ITEMS = 10000;
+                    function fingerprint(el) {
+                        // Prefer id attribute for uniqueness
+                        if (el.id) return 'id:' + el.id;
+                        // Look for data-* ID attributes (data-id, data-item-id,
+                        // data-user-id, etc.) but skip type/role markers like
+                        // data-testid which are shared across items.
+                        for (const attr of el.attributes) {
+                            if (attr.name.startsWith('data-')
+                                && (attr.name === 'data-id'
+                                    || attr.name.endsWith('-id')
+                                    || attr.name.endsWith('-key'))
+                                && attr.value) {
+                                return 'data:' + attr.name + '=' + attr.value;
+                            }
+                        }
+                        // Fallback: two independent 32-bit hashes combined into
+                        // a 64-bit-equivalent string to avoid collisions at scale.
+                        const tag = el.tagName;
+                        const href = el.getAttribute('href') || '';
+                        const text = (el.textContent || '').trim();
+                        let h1 = 0, h2 = 0x9e3779b9;
+                        for (let i = 0; i < text.length; i++) {
+                            const c = text.charCodeAt(i);
+                            h1 = ((h1 << 5) - h1 + c) | 0;
+                            h2 = ((h2 << 7) ^ (h2 >>> 3) ^ c) | 0;
+                        }
+                        return 'hash:' + tag + ':' + href + ':' + h1 + ':' + h2;
+                    }
+                    // Check if element carries a data-* unique identifier
+                    // (data-item-id, data-user-id, etc.). We do NOT check
+                    // el.id here because structural containers commonly have
+                    // ids (id="feed", id="main") without being content items.
+                    function hasDataId(el) {
+                        for (const attr of el.attributes) {
+                            if (attr.name.startsWith('data-')
+                                && (attr.name === 'data-id'
+                                    || attr.name.endsWith('-id')
+                                    || attr.name.endsWith('-key'))
+                                && attr.value) {
+                                return true;
+                            }
+                        }
+                        return false;
+                    }
+                    // Detect whether an element is a repeating-item container
+                    // (a feed/list parent) vs. a content item. A container has
+                    // many children that share the same tagName AND does not
+                    // itself carry a unique identifier.
+                    function isItemContainer(el) {
+                        // Elements with data-*-id are content items, not containers
+                        if (hasDataId(el)) return false;
+                        const kids = el.children;
+                        if (kids.length < 3) return false;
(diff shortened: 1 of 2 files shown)


hafezparast:
Moved changes back to #1853 to keep the review history in one place. Closing this one.

PR #1886, opened 2026-03-30
fix(security): harden _safe_eval_expression with attribute allowlist

ntohidi:
## Summary

- Replaces the blocklist-based attribute validation in `_safe_eval_expression()` with a strict allowlist approach
- Only known-safe attribute/method names are now permitted in computed field expressions
- Blocks lambda expressions, generators, comprehensions, and indirect function calls
- Adds a blocklist for dangerous builtin name references

All legitimate computed field expressions (string methods, type conversions, arithmetic, dict/list access, conditionals) continue to work as expected.

## Test Plan

- [x] 146 unit tests added covering security validations, functional correctness, and edge cases
- [ ] Verify no regressions in existing extraction strategy tests
- [ ] Test with Docker deployment

### DIFF
--- a/crawl4ai/extraction_strategy.py
+++ b/crawl4ai/extraction_strategy.py
@@ -1052,13 +1052,47 @@ _SAFE_EVAL_BUILTINS = {
     "isinstance": isinstance, "type": type,
 }
 
+# Allowlist of attribute names permitted in computed field expressions.
+# Uses allowlist (not blocklist) to prevent sandbox escapes via Python internal
+# attributes like gi_frame, f_back, f_builtins, cr_frame, etc.
+_SAFE_EVAL_ALLOWED_ATTRS = frozenset({
+    # String methods
+    "upper", "lower", "strip", "lstrip", "rstrip", "title", "capitalize",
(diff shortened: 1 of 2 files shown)


unclecode:
Hi @ntohidi -- thank you for the excellent work on this PR! The allowlist approach, the comprehensive test suite (146 tests!), and blocking lambdas/generators/comprehensions were all really well done.

After reviewing both this PR and #1855, we decided to take a hybrid approach that's now landed on develop:

1. **Removed eval() entirely from the computed field `expression` path** -- no sandbox survives long-term against untrusted input (format string bypasses, future Python features, etc). The `function` key still works for SDK users.
2. **Applied your allowlist approach to `_safe_eval_config` in server.py** -- the config deserializer still needs eval, so we hardened it with name allowlists, attribute allowlists, and your lambda/generator/comprehension blocks.
3. **Removed getattr/setattr from hook_manager** -- sandbox escape vectors.
4. **67 adversarial tests** covering all three attack surfaces.

Your allowlist design directly informed the server.py hardening. Really appreciate the thorough security work here.

PR #1891, opened 2026-04-01
fix: scope downloaded_files per crawl to prevent race condition (#1889)

### LINKED ISSUE #1889: [Bug]: `downloaded_files` race condition causes cross-contamination between CrawlResults
## Summary

When `accept_downloads=True` and crawling multiple URLs (via `arun` with a deep crawl strategy, or `arun_many`), the `downloaded_files` list on a `CrawlResult` can contain files triggered by a **previous** URL's navigation. A slow download started by URL A completes after URL B begins processing, causing the file to appear on URL B's result instead of (or in addition to) URL A's.

`AsyncPlaywrightCrawlerStrategy._downloaded_files` is a mutable list on `self`. The lifecycle is:

1. **Reset** at the top of `_crawl_web` (line ~525):
   ```python
   self._downloaded_files = []
   ```

2. **Populated** by a fire-and-forget download listener registered per-page (line ~658):
   ```python
   page.on(
       "download",
       lambda download: asyncio.create_task(
           self._handle_download(download)
       ),
   )
   ```

3. **Read** at the end of `_crawl_web` when building the response (line ~1059):
   ```python
   downloaded_files=(
       self._downloaded_files if self._downloaded_files else None
   ),
   ```

The problem is step 2: `_handle_download` is an `asyncio.create_task` — it runs concurrently. When a download is slow (large file, slow server), the sequence becomes:

```
URL A navigation starts → browser triggers download for URL A
URL A _crawl_web returns → downloaded_files = [] (download still in progress)
URL B _crawl_web starts → self._downloaded_files = [] (reset)
URL A's download completes → _handle_download appends to self._downloaded_files
URL B _crawl_web returns → downloaded_files = [URL_A's_file]  ← WRONG
```

The file that belongs to URL A ends up on URL B's `CrawlResult.downloaded_files`.

In batch/deep-crawl mode the page is reused across URLs, and the download listener's closure captures `self` (not a snapshot of the list), so late-completing downloads always append to whichever `self._downloaded_files` is current at completion time.

### Compounding factor: duplicate file references

Because the list is reset but the listener persists, the same downloaded file can also appear on **multiple** subsequent `CrawlResult` objects if another download event fires (or the listener is re-registered without removing the old one).

## Impact

Any consumer that trusts `CrawlResult.downloaded_files` to identify what document a given URL produced will misattribute content. In our use case (grant document scraping), this caused:

- PDF content attributed to an `.aspx` HTML page
- XLSX scoring template content attributed to a PDF URL
- HTML pages misclassified as binary documents
- Extracted text from the wrong document entirely, silently corrupting downstream data

## Suggested fix

The root cause is that `_downloaded_files` is shared mutable state on `self` with no synchronization between the fire-and-forget download task and the per-URL crawl lifecycle.

### Option A: Scope downloads per-URL with a unique token

Give each `_crawl_web` invocation a unique ID. Pass it into `_handle_download` via the closure. Store downloads in a `dict[str, list[str]]` keyed by invocation ID. Read only the current invocation's list when building the response.

```python
# In _crawl_web:
crawl_id = uuid4().hex
self._downloads_by_crawl[crawl_id] = []

page.on(
    "download",
    lambda download, _id=crawl_id: asyncio.create_task(
        self._handle_download(download, _id)
    ),
)

# ... later:
downloaded_files = self._downloads_by_crawl.pop(crawl_id, []) or None
```

### Option B: Await pending downloads before returning

After page processing completes but before building the `AsyncCrawlResponse`, await any in-flight download tasks. This ensures that downloads triggered by URL A are captured on URL A's result, not URL B's.

```python
# Track download tasks instead of fire-and-forget
task = asyncio.create_task(self._handle_download(download))
self._pending_downloads.append(task)

# Before returning AsyncCrawlResponse:
if self._pending_downloads:
    await asyncio.gather(*self._pending_downloads, return_exceptions=True)
    self._pending_downloads.clear()
```

### Option C: Attach downloads to the navigation request

Associate each download with the navigation that triggered it via Playwright's `Download.page` property and cross-reference with the current page URL. This is the most precise but requires careful handling of redirects.

## Our workaround

We implemented three guards in our post-crawl processing layer:

1. **`seen_downloads` set** — accumulate processed file paths across the stream; skip files already seen.
2. **HTML page guard** — if the page loaded successfully with non-empty markdown, ignore any `downloaded_files` (they're stale noise from a prior URL).
3. **Extension match guard** — reject downloaded files whose extension doesn't match the URL's expected document type (e.g., an `.xlsx` file on a `.pdf` URL).

These heuristics work but are imperfect — they can't handle the case where two consecutive URLs are both documents. A proper fix in crawl4ai would eliminate the need for downstream workarounds entirely.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
1. Configure a crawl with `accept_downloads=True` and a `BestFirstCrawlStrategy` (or any multi-URL crawl).
2. Include at least one URL that triggers a browser download (e.g., a direct link to a large XLSX or PDF — large enough that the download takes >100ms).
3. Include at least one subsequent URL that is a normal HTML page.
4. Observe that the HTML page's `CrawlResult.downloaded_files` contains the file from the previous URL.

Minimal repro sketch:


import asyncio
from crawl4ai import AsyncWebCrawler, BrowserConfig, CrawlerRunConfig

async def main():
    browser_config = BrowserConfig(
        accept_downloads=True,
        downloads_path="/tmp/crawl4ai-downloads",
    )
    run_config = CrawlerRunConfig()

    async with AsyncWebCrawler(config=browser_config) as crawler:
        # URL A: triggers a slow download
        result_a = await crawler.arun("https://example.com/large-file.xlsx", config=run_config)
        # URL B: normal HTML page, crawled immediately after
        result_b = await crawler.arun("https://example.com/normal-page.html", config=run_config)

        print(f"A downloaded_files: {result_a.downloaded_files}")  # Often None (download hadn't finished)
        print(f"B downloaded_files: {result_b.downloaded_files}")  # Contains large-file.xlsx!

asyncio.run(main())


The race is timing-dependent, so it may not reproduce on every run — it depends on download speed relative to the next URL's navigation time. It is **very** reproducible in deep-crawl/streaming mode where URLs are processed back-to-back.
```

### Code snippets

```python

```

### OS

Ubuntu 24.04

### Python version

3.12

### Browser

Chromium

### Browser version

143.0.7499.4

_No response_

hafezparast:
## Summary
- Fixes a race condition where `downloaded_files` from URL A leak into URL B's `CrawlResult` when downloads are slow
- Root cause: `self._downloaded_files` was shared mutable state with fire-and-forget `asyncio.create_task` download handlers — no synchronization between concurrent crawls
- Fix: replace the shared list with a `dict[crawl_id, list]` keyed by a unique ID per `_crawl_web` invocation, passed into `_handle_download` via closure

Fixes #1889

## The Race (before)
```
URL A navigation → browser triggers download
URL A _crawl_web returns → downloaded_files = [] (download still in progress)
URL B _crawl_web starts → self._downloaded_files = [] (reset)
URL A's download completes → appends to self._downloaded_files
URL B _crawl_web returns → downloaded_files = [URL_A's_file] ← WRONG
```

## After Fix
```
URL A gets crawl_id="abc" → downloads scoped to _downloads_by_crawl["abc"]
URL B gets crawl_id="xyz" → downloads scoped to _downloads_by_crawl["xyz"]
Late download for "abc" → goes to "abc"'s list (or logged if already popped)
No cross-contamination possible
```

## Changes
- `crawl4ai/async_crawler_strategy.py`: Replace `self._downloaded_files` (list) with `self._downloads_by_crawl` (dict keyed by crawl_id). `_handle_download` now takes a `crawl_id` parameter. Late downloads that complete after their crawl finished are logged with a warning instead of silently contaminating the next crawl.
- `tests/test_download_race_condition_1889.py`: 9 tests covering isolation, late downloads, concurrent race simulation

## Test plan
- [x] 9 unit tests passing (`pytest tests/test_download_race_condition_1889.py -v`)
- [x] Two crawls with isolated downloads verified
- [x] Late download after crawl finishes — logged, not attached to wrong result
- [x] Concurrent race simulation — slow download on A, fast B, no cross-contamination
- [ ] Integration test with real browser downloads (needs Playwright)

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -50,7 +50,7 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
     Attributes:
         browser_config (BrowserConfig): Configuration object containing browser settings.
         logger (AsyncLogger): Logger instance for recording events and errors.
-        _downloaded_files (List[str]): List of downloaded file paths.
+        _downloads_by_crawl (Dict[str, List[str]]): Downloaded file paths keyed by crawl invocation ID.
         hooks (Dict[str, Callable]): Dictionary of hooks for custom behavior.
         browser_manager (BrowserManager): Manager for browser creation and management.
 
@@ -95,7 +95,7 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
         self.adapter = browser_adapter or PlaywrightAdapter()
 
         # Initialize session management
-        self._downloaded_files = []
+        self._downloads_by_crawl: Dict[str, List[str]] = {}
 
         # Initialize hooks system
         self.hooks = {
@@ -532,8 +532,9 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
         redirected_url = url
         redirected_status_code = None
 
-        # Reset downloaded files list for new crawl
-        self._downloaded_files = []
+        # Scope downloaded files to this crawl invocation
+        crawl_id = uuid.uuid4().hex
+        self._downloads_by_crawl[crawl_id] = []
         
         # Initialize capture lists
         captured_requests = []
@@ -711,12 +712,12 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             if config.fetch_ssl_certificate:
                 ssl_cert = SSLCertificate.from_url(url)
 
-            # Set up download handling
+            # Set up download handling scoped to this crawl invocation
             if self.browser_config.accept_downloads:
                 page.on(
                     "download",
-                    lambda download: asyncio.create_task(
-                        self._handle_download(download)
+                    lambda download, _cid=crawl_id: asyncio.create_task(
+                        self._handle_download(download, _cid)
                     ),
                 )
 
@@ -1155,7 +1156,7 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
                 get_delayed_content=get_delayed_content,
                 ssl_certificate=ssl_cert,
                 downloaded_files=(
-                    self._downloaded_files if self._downloaded_files else None
+                    self._downloads_by_crawl.pop(crawl_id, []) or None
                 ),
                 redirected_url=redirected_url,
                 redirected_status_code=redirected_status_code,
@@ -1454,23 +1455,13 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
             )
             # Continue with normal flow even if virtual scroll fails
 
-    async def _handle_download(self, download):
+    async def _handle_download(self, download, crawl_id: str):
         """
-        Handle file downloads.
-
-        How it works:
-        1. Get the suggested filename.
-        2. Get the download path.
-        3. Log the download.
-        4. Start the download.
-        5. Save the downloaded file.
-        6. Log the completion.
+        Handle file downloads, scoped to a specific crawl invocation.
 
         Args:
             download (Download): The Playwright download object
-
-        Returns:
-            None
+            crawl_id (str): Unique ID of the crawl invocation that triggered this download
         """
         try:
             suggested_filename = download.suggested_filename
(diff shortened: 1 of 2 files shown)


discopops (approved):
Approved. The regression harness change is minimal, the lockfile churn was cleaned, and the race-condition fix validates with `pytest -q tests/test_download_race_condition_1889.py`.

Issue #1256, opened 2025-06-29
[Bug]: Memory Leak on Repeated /md Requests via Docker (MacOS) — Container Crashes Randomly Over Time


Parth7030:
### crawl4ai version

0.6.0-r2 or Latest docker image

### Expected Behavior

When sending repeated POST requests to /md every 10 seconds with dynamic URLs, the Docker container should handle each request efficiently without memory leaks or crashes, maintaining stable memory usage over time.

### Current Behavior

Initially, everything works fine. However, after some time (randomly, depending on duration and frequency), memory usage starts increasing significantly until the container crashes. This suggests a possible memory leak or improper cleanup of resources on repeated executions.

Attached below is a screenshot showing the memory spike before the crash.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
Example request payload sent via Postman every 10 seconds:

{
  "url": "http://www.example.com",
  "f": "fit",
  "q": null,
  "c": "0"
}
```

### Steps to Reproduce

```bash
1.	Run the crawl4ai app using the Docker setup.
2.	Use a tool (like Postman or a script) to send POST requests to http://localhost/md every 10 seconds.
3.	Use a changing URL in the request payload for each call.
4.	Monitor Docker memory usage over time.
5.	Observe the steady increase in memory and eventual container crash.
```

### Code snippets

```python

```

### OS

macOS Sequoia Version 15.5

### Python version

3.12

### Error logs & Screenshots (if applicable)

<img width="1711" alt="Image" src="https://github.com/user-attachments/assets/819c81e7-1c9c-490b-9348-0e8a68cf55ff" />

<img width="1262" alt="Image" src="https://github.com/user-attachments/assets/c804fc52-4457-4568-b482-14e0ce5dcd6c" />

tanushshukla:
yeah I have same issues as well.

emyco:
We are experiencing the same issue and I would like to provide some additional details:

1. I have attached the logs for your review.
2. Notably, when the container is hanging, the `/health` endpoint still returns a status of 200.

Please let me know if you need any further information.

[logs-from-crawl4ai-in-crawl4ai-575d75bc87-bljrk.log](https://github.com/user-attachments/files/21023354/logs-from-crawl4ai-in-crawl4ai-575d75bc87-bljrk.log)

NuoWenLei:
Is there a current workaround for this? I notice this in 0.6.0 as well as in 0.6.3 (built into docker image).

[...]

PR #1906, opened 2026-04-08
feat: AsyncPlasmateCrawlerStrategy — lightweight alternative to Playwright (no Chrome)

dbhurley:
## Summary

Adds `AsyncPlasmateCrawlerStrategy` — a drop-in alternative to `AsyncPlaywrightCrawlerStrategy` using [Plasmate](https://github.com/plasmate-labs/plasmate) instead of Chrome.

Directly addresses:
- **#1256** — Memory leak / crash in Docker from Chrome processes. Plasmate uses ~64MB RAM per session vs ~300MB and leaves no persistent browser process.
- **#1874** — Token usage tracking. Plasmate returns pre-processed content (text / markdown / SOM) instead of raw HTML, cutting upstream token counts by 10-100× before any LLM call.

## What Plasmate is

Open-source Rust browser engine (Apache 2.0). Fetches pages and returns them as Structured Object Model (SOM) — a compact, semantically clean representation with nav, ads, cookie banners, and boilerplate stripped. Install: `pip install plasmate`.

Compression measured across 45 real sites: **17.7× average**, 77× peak. Every token saved before the LLM is a direct cost reduction.

## Drop-in usage

```python
from crawl4ai.async_plasmate_strategy import AsyncPlasmateCrawlerStrategy

strategy = AsyncPlasmateCrawlerStrategy(
    output_format="markdown",   # text | markdown | som | links
    timeout=30,
    fallback_to_playwright=True,  # retry with Playwright for JS-heavy SPAs
)

async with AsyncWebCrawler(crawler_strategy=strategy) as crawler:
    result = await crawler.arun("https://docs.python.org/3/")
    print(result.markdown[:500])
```

## What changed

| File | Change |
|---|---|
| `crawl4ai/async_plasmate_strategy.py` | New `AsyncPlasmateCrawlerStrategy` implementing `AsyncCrawlerStrategy` ABC |
| `crawl4ai/__init__.py` | Export `AsyncPlasmateCrawlerStrategy` |
| `tests/general/test_plasmate_strategy.py` | 20 unit tests (init, cmd building, crawl, fallback, concurrency) |

## Comparison

| | AsyncPlaywrightCrawlerStrategy | AsyncPlasmateCrawlerStrategy |
|---|---|---|
| RAM per session | ~300MB | ~64MB |
| Chrome required | Yes | No |
| Tokens per page (avg) | ~75,000 (raw HTML) | ~4,200 (SOM/text) |
| JS rendering | Yes | No (use `fallback_to_playwright=True`) |
| Install | `playwright install` (~300MB browser) | `pip install plasmate` |
| Persistent process | Yes (browser stays alive) | No (subprocess per fetch) |

## Notes

- No breaking changes — existing `AsyncPlaywrightCrawlerStrategy` usage is untouched
- `fallback_to_playwright=True` makes it safe for mixed static/SPA crawls
- Subprocess runs in asyncio executor — fully non-blocking, safe for concurrent `gather()` calls
- Tested with Python 3.9+

### DIFF
--- a/crawl4ai/__init__.py
+++ b/crawl4ai/__init__.py
@@ -2,6 +2,7 @@
 import warnings
 
 from .async_webcrawler import AsyncWebCrawler, CacheMode
+from .async_plasmate_strategy import AsyncPlasmateCrawlerStrategy
 # MODIFIED: Add SeedingConfig and VirtualScrollConfig here
 from .async_configs import BrowserConfig, CrawlerRunConfig, HTTPCrawlerConfig, LLMConfig, ProxyConfig, GeolocationConfig, SeedingConfig, VirtualScrollConfig, LinkPreviewConfig, MatchMode
 
--- a/crawl4ai/async_plasmate_strategy.py
+++ b/crawl4ai/async_plasmate_strategy.py
@@ -1,0 +1,237 @@
+"""
+AsyncPlasmateCrawlerStrategy — lightweight alternative to AsyncPlaywrightCrawlerStrategy.
+
+Uses Plasmate (https://github.com/plasmate-labs/plasmate) instead of Chrome/Playwright.
+Plasmate is an open-source Rust browser engine that outputs Structured Object Model (SOM)
+instead of raw HTML, using ~64MB RAM per session vs ~300MB and delivering
+10-100x fewer tokens per page — significantly reducing LLM costs.
+
+Install: pip install plasmate
+Docs:    https://plasmate.app
+"""
+
+from __future__ import annotations
+
+import asyncio
+import shutil
+import subprocess
+from typing import Dict, List, Optional
+
+from .async_crawler_strategy import AsyncCrawlerStrategy
+from .async_logger import AsyncLogger
+from .models import AsyncCrawlResponse
+
+_INSTALL_MSG = (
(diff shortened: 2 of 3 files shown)


mshi-hacks:
A Chrome-free crawling strategy is a big deal for containerized environments where Playwright's Chromium binary is a pain to manage — image sizes, security patches, and arm64 compatibility all improve significantly. The drop-in interface matching `AsyncPlaywrightCrawlerStrategy` is the right design choice. One thing I'm curious about: how does Plasmate handle JavaScript-heavy pages that render content via client-side frameworks? Is there a JS execution mode, or is it primarily for static/SSR content?

dbhurley:
Hi @unclecode — bumping this gently three weeks in. Picking up on @mshi-hacks's note above: the chrome-free angle is the bigger story than the integration itself. Container size, ARM64 support, and the security-patch treadmill on Chromium binaries are real pain points for anyone running Crawl4AI on Cloud Run / Lambda / Fly / Pi-class hardware.

Happy to:

- Rebase against latest `main`
- Add anything you'd like to see in the test matrix (specific URLs, behaviors)
- Move bits to a separate `crawl4ai-plasmate` adapter package if keeping the core lean is preferred — the API surface is small enough to ship either way

No rush — just want to make sure it's not stuck behind something I can clear. Thanks for the project.

PR #1907, opened 2026-04-08
fix: narrow arun return type hint

### LINKED ISSUE #1898: [Bug]: arun() and arun_many() type hinting needs fixing
### crawl4ai version

0.8.6

### Expected Behavior

Something like this

Image
And the docstring comments also say that it returns CrawlResult
Image

### Current Behavior

Image

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
RunManyReturn = Union[
    CrawlResultContainer[CrawlResultT],
    AsyncGenerator[CrawlResultT, None]
]

`arun()` is not an async generator, maybe you guys give it generator properties through CrawlResultContainer but the typing needs fixing.
```

### Steps to Reproduce

```bash
I am using Pylance, I tried Pyrefly from meta too.
```

### Code snippets

```python
import asyncio
from crawl4ai import *

async def main():
        result = await crawler.arun(
            url="https://www.nbcnews.com/business",
        )
        print(result.markdown)

if __name__ == "__main__":
    asyncio.run(main())
```

### OS

Windows

### Python version

3.13.11

### Browser

_No response_

### Browser version

_No response_

_No response_

nightcityblade:
Fixes #1898\n\n## Summary\n- annotate AsyncWebCrawler.arun() with the single-result container type it actually returns\n- clarify the docstring so editors understand the backwards-compatible container behavior\n\n## Testing\n- python3 -m py_compile crawl4ai/async_webcrawler.py

### DIFF
--- a/crawl4ai/async_webcrawler.py
+++ b/crawl4ai/async_webcrawler.py
@@ -210,7 +210,7 @@ class AsyncWebCrawler:
         url: str,
         config: CrawlerRunConfig = None,
         **kwargs,
-    ) -> RunManyReturn:
+    ) -> CrawlResultContainer[CrawlResult]:
         """
         Runs the crawler for a single source: URL (web, local file, or raw HTML).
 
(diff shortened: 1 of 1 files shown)


ntohidi:
@nightcityblade Thanks for the contribution! We went with PR #1901 which includes the same fix along with a test file. Closing this one.

PR #1911, opened 2026-04-10
Fix NlpSentenceChunking destroying sentence order

### LINKED ISSUE #1909: bug: NlpSentenceChunking.chunk() uses list(set(sens)) which destroys sentence order
## Bug Description

```python
```

## Problem

## Fix

SamSi0322:
## Summary

`NlpSentenceChunking.chunk()` used `list(set(sens))` to deduplicate sentences, but `set()` is unordered in Python — this destroyed document sentence order, producing arbitrarily shuffled chunks.

**Fix:** Replaced with `list(dict.fromkeys(sens))` which deduplicates while preserving insertion order (Python 3.7+).

Closes #1909

### DIFF
--- a/crawl4ai/chunking_strategy.py
+++ b/crawl4ai/chunking_strategy.py
@@ -86,7 +86,7 @@ class NlpSentenceChunking(ChunkingStrategy):
         sentences = sent_tokenize(text)
         sens = [sent.strip() for sent in sentences]
 
-        return list(set(sens))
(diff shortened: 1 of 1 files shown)


ntohidi:
Thanks for your contribution. While fixing this, I also found that NlpSentenceChunking.__init__() had a broken re-import (from crawl4ai.le.legacy.model_loader import ...) that shadows the working top-level import. So I've made a new PR, #1913

PR #1929, opened 2026-04-21
docker: listen on all addresses

### LINKED ISSUE #1928: [Bug]: docker image rejects connections over IPv6
### crawl4ai version

latest

### Expected Behavior

`curl -v http://localhost:11235/dashboard` should work.

The example in the README should work: https://github.com/unclecode/crawl4ai#quick-test

### Current Behavior

`curl -v http://localhost:11235/dashboard` fails with

    curl: (56) Recv failure: Connection reset by peer

The example fails with:

    requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

Based on the logs produced by crawl4ai, the service only listens on IPv4 addresses by default:

    [2026-04-21 22:15:47 +0000] [8] [INFO] Listening at: http://0.0.0.0:11235 (8)

`curl`, `requests` and any well-behaved client try to use IPv6 by default (this is the correct behaviour as per [rfc8305](https://www.rfc-editor.org/rfc/rfc8305)). The host is reachable, the connection is refused, and the client aborts.

The solution is for the docker container to listen on IPv6 by default. When listening on IPv6, Linux implicitly sets up a listener for IPv4 too, so this would not break compatibility with any setup which relies on IPv4 connectivity.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
n/a
```

### Steps to Reproduce

```bash
curl -v http://localhost:11235/crawl
```

### Code snippets

```python
# This is the example from README.

# Submit a crawl job
response = requests.post(
    "http://localhost:11235/crawl",
    json={"urls": ["http://example.com"], "priority": 10},
)
if response.status_code == 200:
    print("Crawl job submitted successfully.")

if "results" in response.json():
    results = response.json()["results"]
    print("Crawl job completed. Results:")
    for result in results:
        print(result)
else:
    task_id = response.json()["task_id"]
    print(f"Crawl job submitted. Task ID:: {task_id}")
    result = requests.get(f"http://loc
```

### OS

Linux

### Python version

3.14

### Browser

n/a

### Browser version

n/a

_No response_

atomic-carpenter:
## Summary

Listen on all local addresses by default, so that the service is reachable by curl, requests, etc.

Fixes: https://github.com/unclecode/crawl4ai/issues/1928

deploy/docker/supervisord.conf

## How Has This Been Tested?

    podman run -i -p 11235:11235 --name crawl4ai --shm-size=1g unclecode/crawl4ai

Using the example in README.

## Checklist:

### DIFF
--- a/deploy/docker/supervisord.conf
+++ b/deploy/docker/supervisord.conf
@@ -14,7 +14,7 @@ stderr_logfile=/dev/stderr      ; Redirect redis stderr to container stderr
 stderr_logfile_maxbytes=0
 
 [program:gunicorn]
-command=/usr/local/bin/gunicorn --bind 0.0.0.0:11235 --workers 1 --threads 4 --timeout 1800 --graceful-timeout 30 --keep-alive 300 --log-level info --worker-class uvicorn.workers.UvicornWorker server:app
+command=/usr/local/bin/gunicorn --bind [::]:11235 --workers 1 --threads 4 --timeout 1800 --graceful-timeout 30 --keep-alive 300 --log-level info --worker-class uvicorn.workers.UvicornWorker server:app
 directory=/app                  ; Working directory for the app
 user=appuser                    ; Run gunicorn as our non-root user
 autorestart=true


ntohidi:
@atomic-carpenter Thanks for the clear bug report and the fix! The RFC 8305 explanation was very helpful. LGTM — merging to develop.

PR #1933, opened 2026-04-24
fix: preserve sentence order in NlpSentenceChunking.chunk()

### LINKED ISSUE #1909: bug: NlpSentenceChunking.chunk() uses list(set(sens)) which destroys sentence order

hafezparast:
## Problem

`NlpSentenceChunking.chunk()` returned `list(set(sens))`, which:
1. **Destroyed sentence order** — `set()` is unordered; chunks came back in arbitrary order on every run
2. **Silently dropped duplicates** — intentionally repeated sentences were removed

Fixes #1909

## Fix

Remove the `set()` conversion. Return `sens` directly — the list already has duplicates stripped by `sent_tokenize` for the common case, and preserving order is always correct.

```diff
- return list(set(sens))
+ return sens
```

## Test plan
- [ ] `NlpSentenceChunking().chunk("A. B. A.")` returns `["A.", "B.", "A."]` (order preserved, duplicate kept)
- [ ] Chunking a multi-paragraph document returns sentences in document order

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -302,3 +302,7 @@ MEMORY.md
 
 # Handoff files
 HANDOFF-*.md
+
+# Mermaid dev/demo artifacts
+tests/mermaid_comparison*.html
+tests/mermaid_*_comparison.html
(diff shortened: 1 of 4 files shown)


ntohidi:
@hafezparast Thanks for looking into this! This issue has already been fixed in PR #1913 and merged into the `develop` branch. It will be included in the next release. Closing this one.

PR #1943, opened 2026-04-26
fix: preserve .tail text when removing empty elements (#1938)

### LINKED ISSUE #1938: [Bug]: `remove_empty_elements_fast()` drops trailing text when removing empty elements with non-empty .tail
### crawl4ai version

0.8.5

### Expected Behavior

When an element is removed during HTML cleaning, any trailing text attached to that element through .tail should remain in the document.

Removing an empty element should remove only that element, not the normal text that follows it in the parent node.

### Current Behavior

`cleaned_html` can lose normal prose when an empty leaf element is removed by `remove_empty_elements_fast()`.

  This happens when:

  - an element has no own text content
  - the element is selected for removal as “empty”
  - the text following that element is stored in el.tail

  In that case, the element is removed, but its trailing text is dropped also.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
https://arxiv.org/abs/1806.07572 with `remove_overlay_elements=True`. 

The latter will remove the latex related content rendering the DOM element empty. Thats not yet a bug rather a very strict filter. Then the bug can kick in and will drop the now empty element including the .tail
```

### Steps to Reproduce

```bash
run the script you will see the abstract is truncated after "function".
```

### Code snippets

```python
import asyncio
from crawl4ai import AsyncWebCrawler, CacheMode, CrawlerRunConfig


async def main():
    cfg = CrawlerRunConfig(
        remove_overlay_elements=True,
    )

        result = await crawler.arun(
            url="https://arxiv.org/abs/1806.07572",
            config=cfg,
        )

   
    print(result.cleaned_html)


asyncio.run(main())
```

### OS

Linux

### Python version

3.12.12

### Browser

_No response_

### Browser version

_No response_

_No response_

hafezparast:
## Problem

`remove_empty_elements_fast()` was silently dropping `.tail` text when removing empty elements. In lxml, `.tail` is the text that appears *after* an element's closing tag — e.g. the `, some context` in:

```html
<p><small></small>, some context text here</p>
```

When `<small>` is removed because it's empty, the `, some context text here` tail was lost entirely, corrupting the cleaned HTML.

## Fix

Before `parent.remove(el)`, append `el.tail` to the previous sibling's `.tail` (or to `parent.text` if no previous sibling). This is the same pattern already used for `<script>` removal in the same file (lines 730–747).

```python
tail = el.tail
if tail:
    prev = el.getprevious()
    if prev is not None:
        prev.tail = (prev.tail or "") + tail
    else:
        parent.text = (parent.text or "") + tail
parent.remove(el)
```

## Test plan

- [x] 13 targeted unit tests — all pass
- [x] Full regression suite: 303 passed, 1 skipped, 0 failures

Fixes #1938

### DIFF
--- a/crawl4ai/content_scraping_strategy.py
+++ b/crawl4ai/content_scraping_strategy.py
@@ -562,6 +562,15 @@ class LXMLWebScrapingStrategy(ContentScrapingStrategy):
             ):
                 parent = el.getparent()
                 if parent is not None:
+                    # Preserve tail text — in lxml, el.tail is text that appears
+                    # after el's closing tag; removing el would silently drop it.
+                    tail = el.tail
+                    if tail:
+                        prev = el.getprevious()
+                        if prev is not None:
+                            prev.tail = (prev.tail or "") + tail
+                        else:
+                            parent.text = (parent.text or "") + tail
                     parent.remove(el)
 
         return root


ntohidi:
@hafezparast Thank you for your contribution. This issue was resolved in a previous pull request, so I am closing this one.

PR #1969, opened 2026-05-14
fix: route AsyncLogger output to stderr by default (fixes #1968)

### LINKED ISSUE #1968: [Bug] AsyncLogger writes to stdout, breaking MCP stdio transport
### crawl4ai version

0.8.6 (installed via `pip install crawl4ai==0.8.6` as a dep of mcp-crawl4ai 0.3.1).

### Expected Behavior

When crawl4ai is used inside any program that uses **stdout as a structured
data channel** — e.g. an MCP server using the stdio transport — log output
should go to stderr, leaving stdout clean for the host protocol.

### Current Behavior

`crawl4ai/async_logger.py:139` constructs `self.console = Console()` with no
`file=` argument, so `rich.console.Console` defaults to `sys.stdout`.
Every log line — including the progress markers from `url_status()` such as
`[FETCH] ↓ ...`, `https://... | ✓ | ⏱: 1.52s`, `[SCRAPE] ◆`, `[COMPLETE] ●` —
is written to stdout.

When crawl4ai is wrapped by an MCP stdio server (e.g.
[`wyattowalsh/mcp-crawl4ai`](https://github.com/wyattowalsh/mcp-crawl4ai)),
the MCP transport spec requires stdout to contain **only** newline-delimited
JSON-RPC messages. crawl4ai's log lines corrupt the JSON-RPC stream and the
client's `JSONRPCMessage.model_validate_json` raises Pydantic validation
errors for every leaked line:

```
ERROR mcp.client.stdio: Failed to parse JSONRPC message from server
pydantic_core._pydantic_core.ValidationError: 1 validation error for JSONRPCMessage
Invalid JSON: expected value at line 1 column 2 [type=json_invalid,
   input_value='[FETCH]... ↓ ', input_type=str]
...
input_value='https://www.example.com...path/to/page', input_type=str
input_value='age-infrastructure/ | ✓ | ⏱: 1.52s ', input_type=str
input_value='[SCRAPE].. ◆ ', input_type=str
input_value='[COMPLETE] ● ', input_type=str
```

Functionally the tool call still succeeds (the real JSON-RPC response is
also written to stdout and parses fine), but every scrape produces 6–10
spurious ERROR lines in the host's log and risks confusing some MCP
clients into closing the connection.

Note: even when downstream code passes `BrowserConfig(verbose=False)`,
`url_status()` still emits these lines — `verbose=False` only gates a
subset of log calls.

### Root cause

`crawl4ai/async_logger.py`:

```python
from rich.console import Console
...
class AsyncLogger(AsyncLoggerBase):
    ...
    def __init__(self, ...):
        ...
        self.console = Console()        # ← defaults to sys.stdout
```

### Proposed fix

Default the logger console to `sys.stderr`:

```python
import sys
from rich.console import Console
...
self.console = Console(file=sys.stderr)
```

This is the universal convention for library logging and matches what
`logging.StreamHandler` defaults to. Programs that genuinely want logs
on stdout can override by passing a custom `Console` via the existing
constructor.

Alternatively, expose a `stream` / `console` parameter on `AsyncLogger`
so downstream wrappers (mcp-crawl4ai, FastMCP integrations) can force
stderr without monkey-patching.

### Reproduction

1. `pip install crawl4ai==0.8.6 mcp-crawl4ai==0.3.1 mcp`
2. Run any MCP client (e.g. Claude Desktop, mcp-inspector) against
   `crawl4ai_mcp.server` over stdio.
3. Call the `scrape` tool on any URL.
4. Observe `Failed to parse JSONRPC message from server` errors in the
   client log for each scrape, with `input_value` matching crawl4ai's
   progress markers.

### Environment

- crawl4ai 0.8.6
- mcp-crawl4ai 0.3.1
- mcp (python-sdk) 1.x
- Python 3.14 / 3.11
- macOS (also reproduces on Linux per stdio spec)

cgseyhan:
## Problem

AsyncLogger writes its Rich progress output to **stdout** via the default Console() constructor. When crawl4ai is used as an MCP server over stdio (e.g. with mcp-crawl4ai, Claude Desktop, or mcp-inspector), the log lines are injected into the JSON-RPC stream. Clients fail to parse the messages and report:

Failed to parse JSONRPC message from server

This was reported in #1968.

# async_logger.py – before
self.console = Console()   # writes to stdout by default

ich.Console() without arguments writes to stdout. Library log output should go to stderr — this is the POSIX convention and what Python's own logging.StreamHandler does.

# async_logger.py – after
self.console = console if console is not None else Console(stderr=True)

Two changes:
1. **Default to stderr** : Console(stderr=True) writes to sys.stderr, keeping stdout pristine for protocols that use it as a transport.
2. **Injectable console parameter** : callers that want stdout back (or need a custom target) can pass their own Console instance without monkey-patching.

## Testing

Added 	ests/test_async_logger_stderr.py with 7 tests:

| Test | What it checks |
|------|---------------|
| 	est_default_console_writes_to_stderr_not_stdout | capsys.out is empty after logging |
| 	est_default_console_is_stderr | logger.console.file is sys.stderr |
| 	est_custom_console_is_respected | Injected StringIO console receives output |
| 	est_stdout_console_can_be_injected | Legacy stdout behaviour can be restored |
| 	est_no_output_when_verbose_false | erbose=False still suppresses everything |
| 	est_file_logging_still_works | File logging path unaffected |
| 	est_stdout_clean_across_all_log_levels | End-to-end MCP scenario: stdout has only JSON |

All 7 pass ✅

## Backwards Compatibility

- Users who read console.file directly and expect stdout will see stderr instead. This is intentional and correct.
- Users who need the old behaviour can pass console=Console() (or Console(file=sys.stdout)) to the constructor.
- No other public API changes.

### DIFF
--- a/crawl4ai/async_logger.py
+++ b/crawl4ai/async_logger.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 from enum import Enum
 from typing import Optional, Dict, Any, List
 import os
+import sys
 from datetime import datetime
 from urllib.parse import unquote
 from rich.console import Console
@@ -118,6 +119,7 @@ class AsyncLogger(AsyncLoggerBase):
         icons: Optional[Dict[str, str]] = None,
         colors: Optional[Dict[LogLevel, LogColor]] = None,
         verbose: bool = True,
+        console: Optional[Console] = None,
     ):
         """
         Initialize the logger.
@@ -129,6 +131,10 @@ class AsyncLogger(AsyncLoggerBase):
             icons: Custom icons for different tags
             colors: Custom colors for different log levels
             verbose: Whether to output to console
+            console: Optional custom Rich Console instance. Defaults to stderr
+                so that log output does not pollute stdout-based transports
+                such as MCP stdio. Pass ``Console(file=sys.stdout)`` to
+                restore the previous behaviour.
         """
         self.log_file = log_file
         self.log_level = log_level
@@ -136,7 +142,9 @@ class AsyncLogger(AsyncLoggerBase):
         self.icons = icons or self.DEFAULT_ICONS
         self.colors = colors or self.DEFAULT_COLORS
         self.verbose = verbose
-        self.console = Console()
+        # Default to stderr so log lines do not corrupt stdout-based protocols
+        # (e.g. MCP stdio transport, piped shell commands).
+        self.console = console if console is not None else Console(stderr=True)
 
         # Create log file directory if needed
         if log_file:
--- a/tests/test_async_logger_stderr.py
+++ b/tests/test_async_logger_stderr.py
@@ -1,0 +1,156 @@
+"""
+Tests for issue #1968: AsyncLogger must write to stderr by default so that
+stdout-based transports (e.g. MCP stdio) are not corrupted.
+"""
+
+import importlib.util
+import io
+import sys
+from pathlib import Path
+
+import pytest
+from rich.console import Console
+
+# Load async_logger directly without triggering the full crawl4ai __init__
+# (which pulls in many optional deps like aiofiles, OpenSSL, playwright …).
+_spec = importlib.util.spec_from_file_location(
+    "crawl4ai.async_logger",
+    Path(__file__).parent.parent / "crawl4ai" / "async_logger.py",
+)
+_mod = importlib.util.module_from_spec(_spec)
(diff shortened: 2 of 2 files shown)


ntohidi:
Thanks @cgseyhan for the thorough fix and excellent test coverage! This will be included in the next release.

PR #1986, opened 2026-05-26
Fix: AsyncWebCrawler blocks event loop with `scrap`, use `ascrap` instead.

jtgi:
## Summary
`AsyncWebCrawler.arun()` uses synchronous scraping (`scrap`) which blocks the event loop. It should use  asynchronous `ascrap`.

Discovered during production capacity issue.

## Testing
- Already covered: https://github.com/unclecode/crawl4ai/blob/main/tests/test_scraping_strategy.py, no point asserting ascrap called.


jtgi:
@unclecode small fix +1 -1 🙏

PR #1989, opened 2026-05-28
fix(chunking): preserve sentence order in NlpSentenceChunking.chunk()

### LINKED ISSUE #1909: bug: NlpSentenceChunking.chunk() uses list(set(sens)) which destroys sentence order

devteamaegis:
## Summary

`NlpSentenceChunking.chunk()` currently returns `list(set(sens))`. Because Python's
`set` is unordered, the sentences returned by `sent_tokenize` are shuffled into an
arbitrary order and duplicate sentences are silently discarded.

**Before:**
```python
return list(set(sens))   # arbitrary order, deduplicates silently
```

**After:**
```python
return sens              # document order, keeps duplicates
```

## Impact

Any caller that relies on chunks being in the same order as the original text (e.g.
context-window filling, RAG pipeline chunking, sequential LLM prompts) gets
non-deterministic, out-of-order results with the old code.

## Tests

Added `tests/unit/test_nlp_sentence_chunking_order.py` with two tests:
- `test_nlp_sentence_chunking_preserves_order` — verifies sentences come back in
  document order.
- `test_nlp_sentence_chunking_keeps_duplicates` — verifies duplicate sentences are
  not silently dropped.

The test file uses a lightweight stub to avoid importing the full `crawl4ai` package
(which requires `playwright` and `pyOpenSSL` at import time).

Fixes #1909

### DIFF
--- a/crawl4ai/chunking_strategy.py
+++ b/crawl4ai/chunking_strategy.py
@@ -86,7 +86,7 @@ class NlpSentenceChunking(ChunkingStrategy):
         sentences = sent_tokenize(text)
         sens = [sent.strip() for sent in sentences]
 
-        return list(set(sens))
+        return sens
 
 
 # Topic-based segmentation using TextTiling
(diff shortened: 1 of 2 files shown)


ntohidi:
Thanks for your contribution. However, the fix is included in the latest version, 0.8.7.

I will close this PR now.

PR #2035, opened 2026-06-23
Fix _is_nonsense_url dropping trailing-slash language roots (/en/)

jichaowang02-lang:
## Summary

`AsyncUrlSeeder._is_nonsense_url` filters "very short paths", but the check is inconsistent:

```python
if len(path.strip('/')) < 3 and path not in ['/', '/en', '/de', '/fr', '/es', '/it']:
    return True
```

The length test strips slashes (`path.strip('/')`), but the whitelist membership uses the **un-stripped** `path` (which keeps trailing slashes). So the canonical trailing-slash form of a language root is dropped as nonsense:

| url | before | after |
|---|---|---|
| `/en` | kept | kept ✅ |
| `/en/` | **filtered ❌** | kept ✅ |
| `/de/`, `/fr/`, `/es/`, `/it/` | **filtered ❌** | kept ✅ |
| `/ab/`, `/x` (real junk) | filtered | filtered ✅ |

Trailing slashes are the conventional directory form emitted by sitemaps and by `urljoin` normalization, so with `filter_nonsense_urls` enabled (the default) legitimate localized landing pages (`/en/`, `/de/`, …) were silently discarded from seeding results.

## Fix

Compare the slash-stripped path against the whitelist too, so a language root is kept with or without a trailing slash. Genuinely short non-language paths are still filtered.

## Testing

```text
$ pytest tests/unit/test_nonsense_url_language_roots.py -q
11 passed
```

Adds `tests/unit/test_nonsense_url_language_roots.py`; the trailing-slash language-root cases **fail on the current code (5 failed)** and **pass with this fix**, while short-junk/utility URLs stay filtered.

### DIFF
--- a/crawl4ai/async_url_seeder.py
+++ b/crawl4ai/async_url_seeder.py
@@ -1694,8 +1694,12 @@ class AsyncUrlSeeder:
         if any(pattern in url_lower for pattern in ['?print=', '&print=', '/print/', '_print.']):
             return True
         
-        # 12. Very short paths (likely homepage redirects or errors)
-        if len(path.strip('/')) < 3 and path not in ['/', '/en', '/de', '/fr', '/es', '/it']:
+        # 12. Very short paths (likely homepage redirects or errors).
+        # Compare the slash-stripped path against the whitelist too, so the
+        # canonical trailing-slash form of a language root (e.g. "/en/") is
+        # kept just like "/en" instead of being dropped as nonsense.
+        stripped_path = path.strip('/')
+        if len(stripped_path) < 3 and stripped_path not in ['', 'en', 'de', 'fr', 'es', 'it']:
             return True
         
         return False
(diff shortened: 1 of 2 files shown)


chuenchen309 (commented):
Good, precise catch — the stripped-length-vs-unstripped-membership mismatch is a real inconsistency, and the canonical `/en/` form of a language root is exactly the kind of URL a seeder shouldn't be discarding. I reproduced the rule on both sides to confirm the fix is complete and regression-free:

| path | before | after |
|---|---|---|
| `/en` | kept | kept |
| `/en/` `/de/` `/fr/` `/es/` `/it/` | **filtered** | kept ✅ |
| `/` | kept | kept |
| `/about`, `/robots.txt` | kept | kept |
| `/ab/`, `/x` | filtered | filtered (still junk ✅) |

The new whitelist `['', 'en', 'de', 'fr', 'es', 'it']` is exactly the slash-stripped image of the old `['/', '/en', ...]`, so every URL the old code kept is still kept — the change is strictly additive (it only rescues the trailing-slash variants). Nice touch bypassing `__init__` via `__new__` and stubbing `rank_bm25` in the test, matching the sibling unit test.

One pre-existing, out-of-scope note (not this PR's job): uppercase roots like `/EN/` are still filtered, since the whitelist is lowercase — same as before, just flagging in case the maintainers want to normalize case separately later. Nothing blocking; the fix reads correct and complete to me.

PR #2036, opened 2026-06-23
Fix compute_head_fingerprint folding case-only title/meta changes

jichaowang02-lang:
## Summary

`compute_head_fingerprint` lowercases the entire `<head>` before extracting signals:

```python
head_lower = head_html.lower()
title_match = re.search(r'<title[^>]*>(.*?)</title>', head_lower, re.DOTALL)   # value is lowercased
...
match = re.search(pattern, head_lower)                                          # value is lowercased
```

So the captured **values** are lowercased, not just the tag/attribute matching. Two heads that differ only in the **case** of a title or meta value hash to the same fingerprint:

```python
compute_head_fingerprint("<head><title>iPhone</title></head>")
  == compute_head_fingerprint("<head><title>IPHONE</title></head>")   # ❌ same hash
```

`CacheValidator` treats an equal fingerprint as "unchanged", so a genuinely updated page is reported **FRESH** and stale cached content is served.

## Fix

Match tags/attributes case-insensitively with `re.IGNORECASE` against the **original** head, so extracted values keep their case. Tag/attribute case-insensitivity is preserved (e.g. `<TITLE>` / `META NAME=...CONTENT=` still parse), and identical content still hashes identically.

## Testing

```text
$ pytest tests/cache_validation/test_head_fingerprint.py -q
14 passed
```

Adds two regression tests:
- `test_value_case_change_changes_fingerprint` — a case-only value change now changes the fingerprint (**fails on the current code**).
- `test_tag_and_attribute_case_does_not_change_fingerprint` — markup-only case differences still yield the same fingerprint.

### DIFF
--- a/crawl4ai/utils.py
+++ b/crawl4ai/utils.py
@@ -2920,11 +2920,15 @@ def compute_head_fingerprint(head_html: str) -> str:
     if not head_html:
         return ""
 
-    head_lower = head_html.lower()
     signals = []
 
+    # Match tags/attributes case-insensitively, but extract values from the
+    # ORIGINAL head: lowercasing the whole head first would fold a case-only
+    # title/meta change (e.g. "iPhone" -> "IPHONE") to the same fingerprint, so
+    # the cache validator would treat a genuinely changed page as unchanged.
+
     # Extract title
-    title_match = re.search(r'<title[^>]*>(.*?)</title>', head_lower, re.DOTALL)
+    title_match = re.search(r'<title[^>]*>(.*?)</title>', head_html, re.DOTALL | re.IGNORECASE)
     if title_match:
         signals.append(title_match.group(1).strip())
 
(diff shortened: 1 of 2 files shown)


chuenchen309 (commented):
Confirmed the bug against `main` — the analysis is right:

```python
compute_head_fingerprint("<head><title>iPhone</title></head>")
 == compute_head_fingerprint("<head><title>IPHONE</title></head>")   # True on main
compute_head_fingerprint("<head><title>iPhone</title></head>")
 != compute_head_fingerprint("<head><title>Android</title></head>")  # True — so it's the case-folding specifically
```

`head_lower = head_html.lower()` was doing double duty: making the *tag* match case-insensitive (which you want) and lowercasing the captured *value* (which you don't). Splitting those with `re.IGNORECASE` + matching against the original is the right fix, and the meta patterns still match `PROPERTY="OG:TITLE"` style markup the same as before.

**One upgrade effect worth putting in the PR description / release notes.** This changes the fingerprint of essentially every page, not just case-only-changed ones — the extracted signal itself is different now:

```python
head = "<head><title>My Product Page</title></head>"
# main extracts: 'my product page'
# this PR extracts: 'My Product Page'
```

`cache_validator.py:131` and `:163` compare `new_fingerprint == stored_head_fingerprint`, and every stored fingerprint out there was computed by the lowercasing version. So on the first check after upgrading, any cached page whose title has an uppercase letter compares unequal → reported as changed → re-fetched. It's one-time and self-correcting (the new fingerprint gets stored), and it's the safe direction to fail in, but a full cache revalidation right after an upgrade looks a lot like a bug from the outside. Cheap to pre-empt with a line in the release notes.

Not blocking — the fix is correct and the regression test pins the right property. Just flagging so nobody has to debug the revalidation wave.

(Disclosure: I use an AI coding assistant. Both snippets above are ones I ran against `main` myself.)

PR #2046, opened 2026-07-01
Upgrade dependencies

romanbsd:
## Summary
Upgrade dependencies and make some order

## How Has This Been Tested?
pytest

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code


ntohidi:
Thanks for the contribution, @romanbsd.

The `model_loader.py` fix (removing the deprecated `resume_download` kwarg) and the accompanying tests look good. However, the bulk dependency upgrade is too broad to review and merge safely as a single PR.

Could you split this into separate PRs?

1. **model_loader fix + tests** — the `resume_download=None` removal and `test_model_loader_transformers.py`. We can merge this right away.
2. **Patch/minor dependency bumps** — non-breaking version upgrades that keep the existing pinning policy (`~=` where it was).
3. **Major version bumps** — `lxml 5→6`, `numpy 1.26→2.2`, `pillow 10→12`, `chardet 5→7`, etc. These need individual justification and testing since they can introduce breaking API changes.

A few other things to address:
- The pinning policy change from `~=` to `>=` across the board needs maintainer sign-off — please keep existing pin styles unless there's a specific reason to change them.
- `pdf2image` was added as a core dependency — was this intentional? It wasn't previously in the dependency list.
- `setuptools` minimum jumped from 64 to 81 — this could break builds on older systems.

romanbsd:
Thanks for the thorough review, @ntohidi — all fair points. I've split this into three focused PRs against `develop` and closed this one:

1. **#2053 — `model_loader` fix + tests.** Just the `resume_download=None` removal and `test_model_loader_transformers.py`. Ready to merge on its own.
2. **#2054 — minor/patch dependency bumps.** Non-breaking, within-major bumps only.
3. **#2055 — major version bumps.** `lxml 5→6`, `numpy 1.26→2.2`, `pillow 10→12`, `chardet 5→7`, plus `aiofiles`, `snowballstemmer`, `pyOpenSSL`, `psutil`, `rich` — each listed with from→to for individual review.

On your specific concerns:
- **Pinning policy** — reverted. The blanket `~=` → `>=` change is gone; existing `~=` pins are kept as-is (aiosqlite, python-dotenv, requests, beautifulsoup4, xxhash, lxml, snowballstemmer).
- **`pdf2image`** — moved into the `[pdf]` optional-dependencies extra instead of core, since it's only used for PDF processing (#2054).
- **`setuptools` 64 → 81** (and the `wheel` pin) — dropped entirely. Nothing here required it and it risks older-system builds.
- Also dropped the new pins that were added to previously-unpinned optional deps (torch, transformers, etc.).

I also left `uv.lock` out of the two dependency PRs to keep them reviewable — happy to add regenerated lockfiles if you'd prefer them in-PR (note `develop`'s current `uv.lock` is already out of sync with its `pyproject.toml`, e.g. `litellm` vs `unclecode-litellm`).

Issue #1903, opened 2026-04-06
[Bug]: Not Compatible with Python 3.14 (lxml dependency failure)


rayliverified:
### crawl4ai version

0.8.6

## Summary
Crawl4AI installation fails in Python 3.14 environment due to `lxml~=5.3`.
Using `lxml 6.0.2` works!

## AI Analysis

---

`pip install crawl4ai` fails on Windows with Python 3.14 due to `lxml~=5.3` source build error

### Describe the bug

Installing `crawl4ai` on Windows with Python 3.14 fails during dependency installation when `pip` tries to build `lxml` from source.

The failure appears related to Crawl4AI depending on `lxml~=5.3`, which resolves to `lxml 5.4.0` in my environment. On Windows + Python 3.14, `pip` does not appear to find a compatible wheel and falls back to a native source build, which then fails because `libxml2` headers/libraries are not available.

### To Reproduce

1. On Windows, create a fresh virtual environment with Python 3.14
2. Run:

```bash
pip install crawl4ai
```

3. Installation fails while building `lxml`

### Expected behavior

`pip install crawl4ai` should either:

- install successfully on Python 3.14, or
- fail early with a clear message that Python 3.14 is not currently supported

### Actual behavior

Installation fails while building `lxml` from source.

### Error output

```text
Building wheels for collected packages: lxml
  Building wheel for lxml (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for lxml (pyproject.toml) did not run successfully.
  │ exit code: 1

...
fatal error C1083: Cannot open include file: 'libxml/xmlversion.h': No such file or directory
Compile failed: command '...\\cl.exe' failed with exit code 2

...
fatal error C1083: Cannot open include file: 'libxml/xpath.h': No such file or directory

*********************************************************************************
Could not find function xmlXPathInit in library libxml2. Is libxml2 installed?
Is your C compiler installed and configured correctly?
*********************************************************************************

ERROR: Failed building wheel for lxml
Failed to build lxml
error: failed-wheel-build-for-install

× Failed to build installable wheels for some pyproject.toml based projects
╰─> lxml
```

### Environment

- OS: Windows 11
- Python: 3.14.2
- pip: latest available at time of test
- Compiler/tooling: Visual Studio 2022 C++ tools detected in build log

### Additional context

Current Crawl4AI PyPI metadata appears to declare:

- `Requires-Python: >=3.10`
- Python classifiers for:
  - 3.10
[...]

PR #2050, opened 2026-07-03
Fix: Widen lxml dependency ceiling (resolves #2019)fix: widen lxml dependency ceiling to allow lxml>=6.x, fixing scrapli…

### LINKED ISSUE #2019: [Bug]: lxml~=5.3 is too old, can't install crawl4ai & scrapling at the same time
### crawl4ai version

0.8.9

### Expected Behavior

Update the lxml dependency to allow versions >=6.0. This will allow crawl4ai to be used with other lxml-dependent packages like [scrapling](https://github.com/D4Vinci/Scrapling).

### Current Behavior

crawl4ai depends on an old version of the lxml >=5.3,<6.0.
This means crawl4ai cannot be installed along any other dependencies that use lxml>=6.0

One example of that is Scrapling, which is excellent at bypassing Cloudflare bot checks.

Installation with poetry fails due to incompatible lxml dependencies:

```
$ poetry init
...
$ poetry add scrapling
...
$ poetry add crawl4ai
Using version ^0.8.9 for crawl4ai

Updating dependencies
Resolving dependencies... (2.2s)

Because no versions of crawl4ai match >0.8.9,<0.9.0
 and crawl4ai (0.8.9) depends on lxml (>=5.3,<6.0), crawl4ai (>=0.8.9,<0.9.0) requires lxml (>=5.3,<6.0).
And because scrapling (0.4.9) depends on lxml (>=6.1.1)
 and no versions of scrapling match >0.4.9,<0.5.0, crawl4ai (>=0.8.9,<0.9.0) is incompatible with scrapling (>=0.4.9,<0.5.0).
So, because test depends on both scrapling (>=0.4.9,<0.5.0) and crawl4ai (^0.8.9), version solving failed.
```

Installation with pip just replaces lxml with version 5.4.0, which means breaking lxml API for all other packages in the environment.

As a side note, this old lxml dependency causes another issue on Windows: https://github.com/unclecode/crawl4ai/issues/1903

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
See the Current behavior
```

### Code snippets

```python

```

### OS

Any. It's a dependency configuration issue.

### Python version

Python 3.13.13. All versions are affected.

### Browser

_No response_

### Browser version

_No response_

_No response_

mrinal22258:
Fixes #2019

## Root Cause
Currently, `crawl4ai` pins `lxml` to `~=5.3` (resolving to `>=5.3,<6.0`). However, packages commonly co-installed alongside it, such as `scrapling` (used for bypassing bot detection), require `lxml>=6.1.1`.

Because these two ranges have zero overlap:
1. **Poetry** dependency resolution fails outright with `version solving failed`.
2. **Pip** installs `lxml 5.4.0` last, silently breaking whichever package expected `6.x` API compatibility.

In addition, the rigid `<6.0` pin causes build failures on Python 3.14 (`#1903`) where prebuilt wheels for 5.x are not available and compiling from source fails, whereas `lxml>=6.0.2` has been confirmed to work.

## Proposed Changes
We widen the constraint to `lxml>=5.3,<7`. This sets a safe upper ceiling under version 7.0 while allowing `6.x` versions to be successfully resolved.

### Dependency Diff
```diff
# pyproject.toml / requirements.txt
-lxml~=5.3
+lxml>=5.3,<7
```

## Verification & Testing

**Pip Co-installation:** Verified in a clean virtual environment:
```bash
pip install -e .
pip install scrapling
python -c "import lxml; from lxml import etree, html; print(lxml.__version__)"
# Output: 6.1.1 (Successful co-installation and load)
```

**Poetry Resolution:** Verified in a clean Poetry project:
```bash
poetry init
poetry add scrapling
poetry add <local-crawl4ai-path>
# Successfully resolved and wrote lockfile without conflicts.
```

**Test Suite:** Ran the regression extraction strategies and GFM table parsing tests to ensure no APIs broke between lxml 5.x and 6.x:
```bash
pytest tests/regression/test_reg_extraction.py   # 20 passed
pytest tests/test_table_gfm_compliance.py        # 13 passed
```

Note: This change may also resolve or mitigate Python 3.14 build issues described in #1903 by allowing the compiler/resolver to pick a newer lxml wheel, though this has not been independently verified on Python 3.14.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "aiohttp>=3.11.11",
     "aiosqlite~=0.20",
     "anyio>=4.0.0",
-    "lxml~=5.3",
+    "lxml>=5.3,<7",
     "unclecode-litellm==1.81.13",
     "numpy>=1.26.0,<3",
     "pillow>=10.4",
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ aiofiles>=24.1.0
 aiohttp>=3.11.11
 aiosqlite~=0.20
 anyio>=4.0.0
-lxml~=5.3
+lxml>=5.3,<7
 unclecode-litellm==1.81.13
 numpy>=1.26.0,<3
 pillow>=10.4


mrinal22258:
Hi @ntohidi,

I noticed your feedback on the dependency upgrade PR about splitting the changes into smaller PRs.

I already opened this focused PR (#2050), which only widens the `lxml` dependency ceiling to address #2019, with regression tests and dependency resolution verification included.

If this issue is still open and this approach aligns with the project's direction, I'd appreciate it if you could take a look when you have time. If there's anything you'd like changed, I'm happy to update it.

If this is no longer needed or another fix has already been merged, just let me know and I'll close the PR accordingly.

Thanks!

ntohidi:
Thanks for the thorough testing, @mrinal22258 — the regression and co-installation verification helped confirm the change was safe. We've applied the fix directly to `develop` in 057b45b.

PR #2051, opened 2026-07-03
fix: skip channel chromium on Windows to prevent TargetClosedError

fstark96:
## Problem

BrowserConfig.chrome_channel defaults to chromium. When passed to playwright.chromium.launch() as channel=chromium, Playwright looks for a system Chrome installation instead of using the bundled ms-playwright binary.

On Windows this causes Chrome to exit immediately with code 0, resulting in TargetClosedError.

## Fix

Skip passing channel when it equals the default chromium value.


ntohidi:
Thanks for the contribution! Clean fix for a real issue.

fstark96:
> Thanks for the contribution! Clean fix for a real issue.

great that it actually helped you'll
its my first fork lmao
happy to hear

PR #2054, opened 2026-07-04
build(deps): minor/patch dependency bumps

romanbsd:
## Summary

Split out of #2046, per @ntohidi's review. Non-breaking version bumps that stay **within the current major** for each package.

## What changed
- Minor/patch bumps across the core deps (aiohttp, anyio, playwright, patchright, pydantic, cssselect, click, brotli, humanize, lark, shapely, PyYAML, nltk, xxhash, aiosqlite, python-dotenv, requests, beautifulsoup4, fake-useragent, playwright-stealth).
- **Pin styles preserved** — `~=` kept everywhere it was already used (aiosqlite, python-dotenv, requests, beautifulsoup4, xxhash), per your note about keeping the existing pinning policy.
- Consolidated the duplicate `httpx` / `httpx[http2]` entries into a single `httpx[http2]` line.
- Moved `pdf2image` into the `[pdf]` optional-dependencies extra (only needed for PDF processing, not core crawling) rather than making it a core dependency.

## Deliberately excluded (addressing your other points)
- `~=` → `>=` blanket policy change — **reverted**, existing pin styles kept.
- `setuptools` 64 → 81 and the `wheel` pin — **dropped**; nothing here needs them and they risk older-system builds.
- New pins on previously-unpinned optional deps (torch, transformers, etc.) — **dropped**.

## Notes
- `uv.lock` intentionally left out to keep this reviewable; it can be regenerated on merge (develop's committed lock is already stale vs its own `pyproject.toml`).
- Major-version bumps are in a separate PR.


ntohidi:
@romanbsd 
The bumps all check out; minor/patch only, and pinning styles preserved.

One thing: pdf2image was correctly moved to the [pdf] optional extra in pyproject.toml, but it's still listed as an unconditional dependency in requirements.txt. Could you move it under a # [pdf] extra comment section there as well to keep things consistent?

PR #2063, opened 2026-07-10
fix(docker): playground config type + monitor websocket auth dep

### LINKED ISSUE #2059: [Bug]: Docker playground — custom Advanced Config always 400s (/config/dump call omits required `type` field)
### crawl4ai version

0.9.0 (Docker server). Also present on `develop` — see below.

### Summary

In the Docker playground (`/playground`), entering **any** Advanced Config (CrawlerRunConfig or BrowserConfig) and clicking **Run** immediately fails with:

```json
{ "error": "{\"detail\":\"type must be 'CrawlerRunConfig' or 'BrowserConfig'\"}" }
```

The playground's client JS calls `POST /config/dump` with only `{ code }`, but the 0.9.0 endpoint requires a `type` field, so every custom config is rejected with HTTP 400. There is no way to use a custom `CrawlerRunConfig`/`BrowserConfig` from the playground.

### Steps to reproduce

1. Run the 0.9.0 Docker server with a token set (so the UI is reachable).
2. Open `/playground` and expand **▶ Advanced Config** — it defaults to `CrawlerRunConfig(stream=True, cache_mode=CacheMode.BYPASS)`.
3. Click **Run**.

Result: the Response panel shows `type must be 'CrawlerRunConfig' or 'BrowserConfig'`. Reproduces for `stream=True` and `stream=False`, and for the `BrowserConfig` template.

### Root cause

**Client omits `type`** — `deploy/docker/static/playground/index.html`, `pyConfigToJson()`:

```js
const res = await fetch('/config/dump', {
    method: 'POST',
    headers: { 'Content-Type': 'application/json' },
    body: JSON.stringify({ code }),   // <-- no `type`
});
```

**Server requires `type`** — `deploy/docker/server.py`, `_config_from_json()` (called by `POST /config/dump`):

```py
config_type = data.get("type")
if config_type == "CrawlerRunConfig":
    ...
elif config_type == "BrowserConfig":
    ...
else:
    raise ValueError("type must be 'CrawlerRunConfig' or 'BrowserConfig'")
```

The UI already knows the type — it is selected in the `#cfg-type` `<select>` (CrawlerRunConfig / BrowserConfig) that drives the config templates — it simply isn't sent in the request body.

### Evidence

```
# omitting type -> HTTP 400
$ curl -sX POST .../config/dump -H 'Content-Type: application/json' \
    -d '{"code":"CrawlerRunConfig(stream=False, cache_mode=CacheMode.BYPASS)"}'
{"detail":"type must be 'CrawlerRunConfig' or 'BrowserConfig'"}

# including type -> HTTP 200
$ curl -sX POST .../config/dump -H 'Content-Type: application/json' \
    -d '{"type":"CrawlerRunConfig","code":"CrawlerRunConfig(stream=False, cache_mode=CacheMode.BYPASS)"}'
{"type":"CrawlerRunConfig","params":{ ... }}
```

### Fix (one line)

Send the selected type from `#cfg-type` in `pyConfigToJson()`:

```diff
             const res = await fetch('/config/dump', {
                 method: 'POST',
                 headers: { 'Content-Type': 'application/json' },
-                body: JSON.stringify({ code }),
+                body: JSON.stringify({ type: document.getElementById('cfg-type').value, code }),
             });
```

Verified: with this change `/config/dump` returns 200, and the subsequent `POST /crawl` with the resulting `crawler_config` returns a normal batch result (`"success": true`).

### Still present on `develop`

The `develop` branch (which has the merged auth-gate fix #2039) still omits `type` in the same call — it uses `authFetch('/config/dump', { ... body: JSON.stringify({ code }) })` (~line 569) — so this bug is not fixed there.

### Secondary symptom (same broken path, FYI)

Because the 400 is caught and the fallback sets `advConfig = { crawler_config: { stream: true } }`, while `shouldUseStream()` checks `crawler_config.params.stream` (different nesting), the default `stream=True` template makes the client send a streaming request but parse it as batch. The server then returns NDJSON (`{…result…}\n{"status":"completed"}`) and `response.json()` throws `SyntaxError: Unexpected non-whitespace character after JSON … (line 2 column 1)`. Fixing the `type` omission above resolves the custom-config 400; the `shouldUseStream` nesting mismatch looks like a related follow-up.

### LINKED ISSUE #2060: [Bug]: Docker dashboard WebSocket /monitor/ws 500s — HTTP-only token_dep applied to the WS route (TypeError: _principal() missing 'request')
### crawl4ai version

0.9.0 (Docker server). Also present on `develop` — see below.

### Summary

The monitor dashboard (`/dashboard`) never connects. Its WebSocket to `/monitor/ws` fails immediately; the browser console loops `WebSocket connection to 'ws://<host>/monitor/ws' failed` / `WebSocket closed`, and the header stays on **Reconnecting…**. The server returns **HTTP 500** on the WS upgrade — even on a correctly authenticated request.

### Root cause

`deploy/docker/server.py` includes the monitor router with a router-level dependency:

```py
app.include_router(monitor_router, dependencies=[Depends(token_dep)])
```

`token_dep` is HTTP-only — `deploy/docker/auth.py`:

```py
def get_token_dependency(config: Dict):
    def _principal(request: Request) -> Optional[Dict]:
        return get_principal(request)
    return _principal
```

Because the dependency is attached at the router level, it is applied to **every** route in the router, including the `@router.websocket("/ws")` endpoint. On a WebSocket scope FastAPI cannot inject `request: Request`, so `_principal()` is called with no arguments:

```
TypeError: get_token_dependency.<locals>._principal() missing 1 required positional argument: 'request'
```

...which surfaces as a 500 and the WS never upgrades. Note the WS handler itself declares no `Depends` and its own docstring says auth is handled by the middleware:

```py
@router.websocket("/ws")
async def websocket_endpoint(websocket: WebSocket):
    # Auth is enforced by the AuthGateMiddleware (outermost ASGI layer) ...
    # No bespoke check here.
    await websocket.accept()
```

So the router-level `token_dep` is both **broken for WS** and **redundant** (authentication is already enforced by `AuthGateMiddleware`, which is also what populates `request.state.principal` that `require_admin` reads).

### Traceback

```
File ".../fastapi/dependencies/utils.py", line 680, in solve_dependencies
File ".../starlette/concurrency.py", line 34, in run_in_threadpool
...
TypeError: get_token_dependency.<locals>._principal() missing 1 required positional argument: 'request'
```

### Steps to reproduce

1. Run the 0.9.0 Docker server with `CRAWL4AI_API_TOKEN` set.
2. Open `/dashboard` (send the token however your deployment allows — e.g. reverse-proxy header injection).
3. The dashboard shows **Reconnecting…**; the server logs the `TypeError` above on each `/monitor/ws` attempt (HTTP 500).

(There is a second, independent bug on this same route: `prometheus_fastapi_instrumentator` crashing on FastAPI's `_IncludedRouter` — reported in #2025 — which must also be resolved for the WS to work. This issue is specifically about the `token_dep`-on-WS `TypeError`.)

### Fix

Drop the router-level dependency (auth stays enforced by `AuthGateMiddleware`; the destructive `/monitor/actions/*` and `/monitor/stats/reset` routes keep their own `Depends(require_admin)`):

```diff
-app.include_router(monitor_router, dependencies=[Depends(token_dep)])
+app.include_router(monitor_router)
```

(Alternatively, make the auth dependency WebSocket-aware, e.g. accept `websocket: WebSocket` in addition to `request: Request` — but since the middleware already gates the route, removing the redundant dependency is simplest.)

Verified with the change: `/monitor/ws` upgrades (`101 Switching Protocols`) and streams the expected JSON (`health`, `requests`, `browsers`, `timeline`, `janitor`, `errors`, `timestamp`); an unauthenticated WS is still rejected by `AuthGateMiddleware`; the admin action routes still require an admin credential.

### Still present on `develop`

`develop` has the same line — `app.include_router(monitor_router, dependencies=[Depends(token_dep)])` (server.py, ~line 486) with `_principal(request: Request)` (auth.py, ~line 130) — so this is not fixed there.

Pitchfork-and-Torch:
## Summary

Two Docker UI server bugs with confirmed root causes on `develop`.

### Fixes #2059 — Playground Advanced Config always 400

`pyConfigToJson()` POSTed `{ code }` to `/config/dump`, but the server requires `type` (`CrawlerRunConfig` | `BrowserConfig`). Every custom Advanced Config failed with:

```json
{"detail":"type must be 'CrawlerRunConfig' or 'BrowserConfig'"}
```

**Change:** send `type` from `#cfg-type` with the code.

Also tightens the related stream-path follow-up from the issue:

- Fallback minimal config uses dump-shaped `{ type, params: { stream: true } }`
- `shouldUseStream()` reads both `crawler_config.params.stream` and top-level `crawler_config.stream`

### Fixes #2060 — Dashboard `/monitor/ws` 500

```python
app.include_router(monitor_router, dependencies=[Depends(token_dep)])
```

`token_dep` is HTTP `Request`-only. FastAPI applies router-level deps to the WebSocket route; inject fails with:

```text
TypeError: _principal() missing 1 required positional argument: 'request'
```

The WS handler already documents that `AuthGateMiddleware` owns auth; destructive routes keep `Depends(require_admin)`.

**Change:** `app.include_router(monitor_router)` without the HTTP-only dependency.

## Test plan

- [ ] Playground: expand Advanced Config (default `CrawlerRunConfig`), Run → `/config/dump` returns 200 and crawl succeeds
- [ ] Playground: switch type to `BrowserConfig` template → dump + run succeed
- [ ] Playground: intentional bad Python in Advanced Config still surfaces config error
- [ ] Dashboard: with valid token, `/monitor/ws` upgrades (101) and streams health/requests JSON
- [ ] Dashboard: unauthenticated WS still rejected by middleware
- [ ] Monitor admin actions still require admin scope

## Notes

- Targets `develop` (collaborator convention).
- Minimal, focused diff: 2 files.

### DIFF
--- a/deploy/docker/server.py
+++ b/deploy/docker/server.py
@@ -482,8 +482,12 @@ def _config_from_json(data: dict) -> dict:
 app.include_router(init_job_router(redis, config, token_dep))
 
 # ── monitor router ──────────────────────────────────────────
+# Do not attach token_dep at router level: it is HTTP Request-only and breaks
+# the WebSocket upgrade on /monitor/ws (TypeError: _principal() missing 'request').
+# AuthGateMiddleware already authenticates HTTP + WS; destructive monitor
+# actions keep their own Depends(require_admin).
 from monitor_routes import router as monitor_router
-app.include_router(monitor_router, dependencies=[Depends(token_dep)])
+app.include_router(monitor_router)
 
 logger = logging.getLogger(__name__)
 
--- a/deploy/docker/static/playground/index.html
+++ b/deploy/docker/static/playground/index.html
@@ -563,10 +563,13 @@
             const code = cm.getValue().trim();
             if (!code) return {};
 
+            // Server requires `type` alongside `code` (CrawlerRunConfig | BrowserConfig).
+            // The UI already tracks this in #cfg-type; omitting it always 400s Advanced Config.
+            const cfgType = document.getElementById('cfg-type').value;
             const res = await authFetch('/config/dump', {
                 method: 'POST',
                 headers: { 'Content-Type': 'application/json' },
-                body: JSON.stringify({ code }),
+                body: JSON.stringify({ type: cfgType, code }),
             });
 
             const statusEl = document.getElementById('cfg-status');
@@ -652,9 +655,12 @@
         // Detect if stream is requested inside payload
         function shouldUseStream(payload) {
             const toBool = (v) => v === true || (typeof v === 'string' && v.toLowerCase() === 'true');
-            const fromCrawler = payload && payload.crawler_config && payload.crawler_config.params && payload.crawler_config.params.stream;
+            // Successful /config/dump returns { type, params: { stream, ... } }
+            const fromCrawlerParams = payload && payload.crawler_config && payload.crawler_config.params && payload.crawler_config.params.stream;
+            // Fallback / minimal shapes may set stream at the crawler_config top level
+            const fromCrawlerTop = payload && payload.crawler_config && payload.crawler_config.stream;
             const direct = payload && payload.stream;
-            return toBool(fromCrawler) || toBool(direct);
+            return toBool(fromCrawlerParams) || toBool(fromCrawlerTop) || toBool(direct);
         }
 
         // Main run function
@@ -676,8 +682,8 @@
                 const streamFlag = /stream\s*=\s*True/i.test(codeText);
                 const isCrawlEndpoint = document.getElementById('endpoint').value === 'crawl';
                 if (isCrawlEndpoint && streamFlag) {
-                    // Fallback: proceed with minimal config only for stream
-                    advConfig = { crawler_config: { stream: true } };
+                    // Fallback: minimal dump-shaped config so shouldUseStream + server load agree
+                    advConfig = { crawler_config: { type: 'CrawlerRunConfig', params: { stream: true } } };
                 } else {
                     updateStatus('error');
                     document.querySelector('#response-content code').textContent =


ntohidi:
Thanks for the fix, @Pitchfork-and-Torch.

PR #2067, opened 2026-07-10
fix(docker): include Playwright headless shell

### LINKED ISSUE #2064: [Bug]: The docker image does not launch since 0.9.1
### crawl4ai version

0.9.1

### Expected Behavior

Following the install guide:
```
docker run -d \
  -p 11235:11235 \
  --name crawl4ai \
  --shm-size=1g \
  unclecode/crawl4ai:latest
```

should work.

### Current Behavior

However it repeatedly crashes, downgrading to 0.9.0's image works fine.

Issue seems to be:
```
playwright._impl._errors.Error: BrowserType.launch: Executable doesn't exist at /home/appuser/.cache/ms-playwright/chromium_headless_shell-1228/chrome-headless-shell-linux64/chrome-headless-shell
```

where it's looking for the headless-shell. However in your Dockerfile it's only copying the full chrome binary, so it doesn't exist, and crashes.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python

```

### OS

Linux

### Python version

Whatever version is bundled inside the docker image

### Browser

_No response_

### Browser version

_No response_

I am running podman instead of docker, but I believe the same issue would occur in docker as downgrading to 0.9.0 works fine.

reallav0:
## Summary
- copy Playwright's `chromium_headless_shell-*` cache into the runtime user's cache
- add a Dockerfile regression check for both Chromium artifacts

## Root cause
The image runs as `appuser`, but only copied the full Chromium cache from root. Headless crawls resolve `chromium_headless_shell-*`, so the Docker server exited at startup when that directory was missing.

Fixes #2064.

## Validation
- `git diff --check`
- standard-library Dockerfile regression check passed

Not run: Docker image build (Docker daemon unavailable); focused pytest could not be installed because the full project dependency set exhausted available disk space.

### DIFF
--- a/deploy/docker/tests/test_security_container_posture.py
+++ b/deploy/docker/tests/test_security_container_posture.py
@@ -57,6 +57,19 @@ class TestDockerfile:
         assert "/var/lib/crawl4ai/outputs" in dockerfile
         assert "chmod 700 /var/lib/crawl4ai/outputs" in dockerfile
 
+    def test_playwright_headless_shell_is_copied_to_runtime_cache(self, dockerfile):
+        # Playwright launches chromium_headless_shell for headless crawls. The
+        # image runs as appuser, so both Chromium artifacts must be copied out
+        # of root's install cache.
+        cache_copy = re.search(
+            r"RUN mkdir -p /home/appuser/\.cache/ms-playwright(?P<commands>.*?)RUN crawl4ai-doctor",
+            dockerfile,
+            re.DOTALL,
+        )
+        assert cache_copy, "Dockerfile must copy Playwright artifacts into appuser's cache"
+        assert "chromium-*" in cache_copy.group("commands")
+        assert "chromium_headless_shell-*" in cache_copy.group("commands")
+
     def test_runs_as_non_root(self, dockerfile):
         assert re.search(r"^USER\s+appuser", dockerfile, re.MULTILINE)
 
--- a/deploy/docker/tests/test_security_container_posture.py
+++ b/deploy/docker/tests/test_security_container_posture.py
@@ -62,7 +62,7 @@ class TestDockerfile:
         # image runs as appuser, so both Chromium artifacts must be copied out
         # of root's install cache.
         cache_copy = re.search(
-            r"RUN mkdir -p /home/appuser/\.cache/ms-playwright(?P<commands>.*?)RUN crawl4ai-doctor",
+            r"cp -r(?P<artifacts>(?:(?!&&).)*)/home/appuser/\.cache/ms-playwright/",
             dockerfile,
             re.DOTALL,
         )
--- a/Dockerfile
+++ b/Dockerfile
@@ -178,7 +178,9 @@ RUN crawl4ai-setup
 RUN playwright install --with-deps
 
 RUN mkdir -p /home/appuser/.cache/ms-playwright \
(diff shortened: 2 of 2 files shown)


deploy/docker/tests/test_security_container_posture.py
```diff
@@ -57,6 +57,19 @@ def test_artifact_dir_created_0700(self, dockerfile):
         assert "/var/lib/crawl4ai/outputs" in dockerfile
         assert "chmod 700 /var/lib/crawl4ai/outputs" in dockerfile
 
+    def test_playwright_headless_shell_is_copied_to_runtime_cache(self, dockerfile):
+        # Playwright launches chromium_headless_shell for headless crawls. The
+        # image runs as appuser, so both Chromium artifacts must be copied out
+        # of root's install cache.
+        cache_copy = re.search(
+            r"RUN mkdir -p /home/appuser/\.cache/ms-playwright(?P<commands>.*?)RUN crawl4ai-doctor",
```
SohamKukreti: Consider anchoring the regex on the `cp … /home/appuser/.cache/ms-playwright/` command rather than the region between `mkdir` and `crawl4ai-doctor`, as written, reordering the `Dockerfile` or renaming the doctor step fails this test even when the copy is intact.

SohamKukreti (commented):
Thanks for the quick fix! I dug into #2064 independently and can confirm both the root cause and that this fix works — verified against the published images.

**Root cause (confirmed):**
- `playwright install --with-deps` installs two Chromium artifacts in root's cache: `chromium-1228` and `chromium_headless_shell-1228`. The old copy glob `chromium-*` (hyphen) never matched the underscore-named headless shell, so it's missing from appuser's cache in **both** the 0.9.0 and 0.9.1 images.
- This was harmless until 0.9.1: commit c34ffc0 (#2051) stopped passing `channel="chromium"` for the default config, and with no channel, Playwright ≥1.49 launches headless mode with the headless shell — the one binary the runtime user never had. Boot crash → restart loop.

**Fix verification:** I injected `chromium_headless_shell-1228` into `/home/appuser/.cache/ms-playwright/` in a stock `unclecode/crawl4ai:0.9.1` container (byte-for-byte what this PR's copy produces) and the container boots cleanly — `Application startup complete`, `/health` returns OK, and the permanent browser pool runs with a `chrome-headless-shell` process. Image size cost is +262 MB uncompressed (~7% on a ~3.9 GB image).

Two things this PR gets right that are worth preserving in review:
- **Keeping both binaries** (not just the headless shell): full `chromium-*` is still required for `headless=false` requests — the image ships Xvfb precisely for headed mode. Copying only the shell would fix the boot crash but break headed crawls at request time.
- **Keeping `cp` rather than `mv`**: `crawl4ai-doctor` runs as root after this line and needs root's copy intact.

One non-blocking nit on the test regex (inline). LGTM otherwise, happy to see this un-drafted and merged.

deploy/docker/tests/test_security_container_posture.py
reallav0: Got it bud 🗿

ntohidi:
Thanks for the fix, @reallav0. Merged — will fix the test group name typo in a follow-up commit.

PR #2080, opened 2026-07-15
fix(deep_crawling): allow single-label hostnames

### LINKED ISSUE #2079: [Bug]: BFSDeepCrawlStrategy.can_process_url() rejects valid single-label hostnames (netloc without a dot)
### crawl4ai version

0.9.2

### Expected Behavior

Single-label/internal hostnames without a dot should be treated as valid
hostnames, at least when the URL has a valid scheme (http/https) and
non-empty netloc. The dot-check appears to be an overly strict heuristic
that isn't part of the documented API/config (no `allow_single_label_hosts`
or similar flag exists).

### Current Behavior

- Depth 0 (start URL) is fetched successfully.
- Every discovered link at depth > 0 is logged as:
  `Invalid URL: https://name/xyz/xyz, error: Invalid domain`
  The configured `DomainFilter` is never even reached, since the exception
  is raised before `filter_chain.apply()` is called.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
1. Start URL: `https://name/xyz/` (a host with no dot in its name,
   reachable e.g. via internal DNS or /etc/hosts).
2. Configure `BFSDeepCrawlStrategy` with
   `filter_chain=FilterChain([DomainFilter(allowed_domains=["iltis3"])])`.
3. Run `crawler.arun()`.
```

### Code snippets

```python

```

### OS

Linux

### Python version

3.12.13

### Browser

_No response_

### Browser version

_No response_

[INIT].... → Crawl4AI 0.9.2 

[FETCH]... ↓ https://name/xyz                                                                                   | ✓ | ⏱: 0.91s 

[SCRAPE].. ◆ https://name/xyz                                                                                   | ✓ | ⏱: 0.06s 

[EXTRACT]. ■ https://name/xyz                                                                                   | ✓ | ⏱: 0.03s 

[COMPLETE] ● https://name/xyz                                                                                   | ✓ | ⏱: 1.01s 

Invalid URL: https://name/xyz/neu-im-intranet, error: Invalid domain

Invalid URL: name/xyz/suche, error: Invalid domain

Invalid URL: https://name/xyz

[crawler.py](https://github.com/user-attachments/files/30046146/crawler.py)

/allgemein/stellen, error: Invalid domain

nightcityblade:
## Summary

Fixes #2079.

Allow `BFSDeepCrawlStrategy` to process valid HTTP(S) URLs whose host is a single label, such as internal DNS names. The existing scheme and non-empty netloc validation continues to reject malformed URLs.

- `crawl4ai/deep_crawling/bfs_strategy.py` - Remove the overly strict requirement that every hostname contain a dot.
- `tests/deep_crawling/test_deep_crawl_resume.py` - Add a regression test proving a single-label hostname reaches and passes the configured domain filter.

## How Has This Been Tested?

- `.venv/bin/pytest -q tests/deep_crawling/test_deep_crawl_resume.py` — 33 passed.
- `.venv/bin/black --check crawl4ai/deep_crawling/bfs_strategy.py tests/deep_crawling/test_deep_crawl_resume.py` — run; the command reports existing whole-file formatting drift also present on the upstream baseline. The changed lines were reviewed against Black's diff and introduce no new formatting changes.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas — N/A; the validation removal is self-explanatory.
- [ ] I have made corresponding changes to the documentation — N/A; this is a narrow bug fix with no public API change.

### DIFF
--- a/crawl4ai/deep_crawling/bfs_strategy.py
+++ b/crawl4ai/deep_crawling/bfs_strategy.py
@@ -70,8 +70,6 @@ class BFSDeepCrawlStrategy(DeepCrawlStrategy):
                 raise ValueError("Missing scheme or netloc")
             if parsed.scheme not in ("http", "https"):
                 raise ValueError("Invalid scheme")
-            if "." not in parsed.netloc:
-                raise ValueError("Invalid domain")
         except Exception as e:
             self.logger.warning(f"Invalid URL: {url}, error: {e}")
             return False
--- a/tests/deep_crawling/test_deep_crawl_resume.py
+++ b/tests/deep_crawling/test_deep_crawl_resume.py
@@ -629,6 +629,14 @@ class TestBFSRegressions:
         assert await strategy.can_process_url("https://other.com/blog/post1", 1) == False
 
     @pytest.mark.asyncio
+    async def test_single_label_hostname_is_valid(self):
+        """Single-label hostnames can reach the configured filter chain."""
+        filter_chain = FilterChain([DomainFilter(allowed_domains=["intranet"])])
+        strategy = BFSDeepCrawlStrategy(max_depth=1, filter_chain=filter_chain)
+
+        assert await strategy.can_process_url("https://intranet/docs", 1) is True
+
+    @pytest.mark.asyncio
     async def test_url_scorer_still_works(self):
         """URL scoring integration unchanged."""
         scorer = KeywordRelevanceScorer(keywords=["python", "tutorial"], weight=1.0)


keepitdigital:
k

chuenchen309 (commented):
Reviewed against `can_process_url` — this looks correct and safe.

The removed `"." not in parsed.netloc` check was never the domain-restriction mechanism: right below it, `if depth != 0 and not await self.filter_chain.apply(url)` is what actually enforces `DomainFilter`/allowed-domains, and that is untouched. So the dot-check was purely a crude structural guard that wrongly rejected valid absolute URLs with a single-label host — not just intranet names but also `http://localhost`, `http://localhost:8080`, and Docker/K8s service hostnames, all of which have a non-empty netloc and a valid http(s) scheme.

Points that make this safe:
- **No domain/isolation regression:** domain scoping still runs through `filter_chain.apply()`; single-label hosts now merely *reach* that filter (exactly what the new `DomainFilter(allowed_domains=["intranet"])` test asserts) instead of being dropped before it.
- **Malformed input still rejected:** relative paths / sch/netloc-less strings still fail the existing `not parsed.scheme or not parsed.netloc` guard (empty netloc → rejected), so nothing new slips through.
- **Unfiltered crawls unchanged in spirit:** with no `DomainFilter` configured, dotted hosts were already crawled unrestricted; single-label hosts just join them — the user opted into that by not setting a filter.

The regression test is well-targeted (proves the host both reaches and passes the configured filter). LGTM.

PR #2095, opened 2026-07-22
fix(docker): remove build artifacts from image layers

### LINKED ISSUE #2092: [Bug]: arm64 image ships ~2.4GB of build-time waste (core dumps + duplicate Chromium cache), root-caused to QEMU CI emulation
### crawl4ai version

0.9.2

### Expected Behavior

The published arm64 Docker image should only contain files needed at runtime. Build/test artifacts (crash dumps from the crawl4ai-doctor sanity check) and intermediate install artifacts (the root-owned Playwright browser cache, superseded by the copy made for appuser) should be cleaned up before the image layers are committed.

### Current Behavior

`unclecode/crawl4ai:latest` (arm64) reports as 9.06GB via docker images. Inspecting a freshly-pulled container shows two separate sources of dead weight, totaling ~2.4GB of its ~5.5GB live filesystem:

1. Core dumps in /app (~926MB)

```sh
-r--r--r-- root root 510M /app/qemu_headless_shell_20260715-085214_28.core
-r--r--r-- root root 208M /app/qemu_headless_shell_20260715-085211_34.core
-r--r--r-- root root 208M /app/qemu_headless_shell_20260715-085211_32.core
```

- These are dated at image build time, not container runtime, and are root-owned/read-only, so end users can't remove them from a running container.

2. Duplicate Chromium install (~1.5GB)

```sh
1.5G  /root/.cache/ms-playwright       (dead - root never runs; container runs as appuser)
958M  /home/appuser/.cache/ms-playwright  (actually used at runtime)
```

- `playwright install --with-deps` installs the browser to `/root/.cache` (running as root during build), then a later `RUN step cp -r`'s it to `/home/appuser/.cache` for the actual runtime user, but never deletes the original. Both copies ship in the final image.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
N/A — not input-dependent. Present in every pull of the arm64 image, not triggered by user config.
```

### Steps to Reproduce

```bash
Steps to Reproduce:
1. docker pull unclecode/crawl4ai:latest
2. docker run --rm -it unclecode/crawl4ai:latest bash
3. ls -lh /app/*.core
4. du -sh /root/.cache/ms-playwright /home/appuser/.cache/ms-playwright
```

### Code snippets

```python
docker history — confirms duplicate layer cost:

$ docker history --no-trunc unclecode/crawl4ai:latest | grep -E "chown -R root|cp -r.*ms-playwright"
972MB   RUN chown -R root:root ${APP_HOME} && chmod -R a-w ${APP_HOME}

1GB     RUN mkdir -p /home/appuser/.cache/ms-playwright \
          && cp -r /root/.cache/ms-playwright/chromium-* \
                   /root/.cache/ms-playwright/chromium_headless_shell-* \
                   /home/appuser/.cache/ms-playwright/ \
          && chown -R appuser:appuser /home/appuser/.cache/ms-playwright



docker-release.yml — confirms cross-arch QEMU build:


$ grep -A3 "platforms:" .github/workflows/docker-release.yml

          platforms: linux/amd64,linux/arm64
          cache-from: type=gha
          cache-to: type=gha,mode=max
```

### OS

Linux (Raspberry Pi OS, arm64

### Python version

3.12

### Browser

Chromium (bundled headless_shell, via Playwright build 1228)

### Browser version

149.0.7827.0

docker history shows the `chown -R root:root $APP_HOME && chmod -R a-w $APP_HOME` layer at 972MB, matching the ~926MB of .core files in /app. `du -sh` on a running container shows `/root/.cache/ms-playwright` (1.5G) and `/home/appuser/.cache/ms-playwright` (958M) as separate, duplicate browser installs

nightcityblade:
## Summary

Fixes #2092.

Keep Playwright installation, runtime-cache copying, and root-cache cleanup in one Docker layer so the superseded root-owned browser cache is not committed to the image. Remove QEMU core dumps immediately after `crawl4ai-doctor`, before that layer is committed.

- `Dockerfile` - clean duplicate Playwright data and doctor core dumps in the same layers that create them.
- `deploy/docker/tests/test_security_container_posture.py` - enforce both same-layer cleanup guarantees.

## How Has This Been Tested?

- `docker buildx build --check --progress plain .` — check complete, no warnings found.
- `.venv/bin/python -m pytest deploy/docker/tests/test_security_*.py -q` — 317 passed, 1 xfailed.
- `.venv/bin/black --check --target-version py312 deploy/docker/tests/test_security_container_posture.py` — passed.
- `.venv/bin/ruff check deploy/docker/tests/test_security_container_posture.py` — passed.

A full multi-architecture image build was not run locally; the focused posture test verifies that each cleanup remains in the Docker layer that creates its artifact.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas — N/A; the cleanup commands and focused assertions are self-explanatory.
- [ ] I have made corresponding changes to the documentation — N/A; this changes image contents without changing user-facing behavior.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -175,15 +175,16 @@ RUN pip install --no-cache-dir --upgrade pip && \
 
 RUN crawl4ai-setup
 
-RUN playwright install --with-deps
-
-RUN mkdir -p /home/appuser/.cache/ms-playwright \
+RUN playwright install --with-deps \
+    && mkdir -p /home/appuser/.cache/ms-playwright \
     && cp -r /root/.cache/ms-playwright/chromium-* \
         /root/.cache/ms-playwright/chromium_headless_shell-* \
         /home/appuser/.cache/ms-playwright/ \
-    && chown -R appuser:appuser /home/appuser/.cache/ms-playwright
+    && chown -R appuser:appuser /home/appuser/.cache/ms-playwright \
+    && rm -rf /root/.cache/ms-playwright
 
-RUN crawl4ai-doctor
+RUN crawl4ai-doctor \
+    && find "${APP_HOME}" -maxdepth 1 -type f -name '*.core' -delete
 
 # Ensure all cache directories belong to appuser
 # This fixes permission issues with .cache/url_seeder and other runtime cache dirs
--- a/Dockerfile
+++ b/Dockerfile
@@ -173,17 +173,15 @@ RUN pip install --no-cache-dir --upgrade pip && \
     python -c "import crawl4ai; print('✅ crawl4ai is ready to rock!')" && \
     python -c "from playwright.sync_api import sync_playwright; print('✅ Playwright is feeling dramatic!')"
 
-RUN crawl4ai-setup
-
-RUN playwright install --with-deps \
+RUN crawl4ai-setup \
+    && playwright install --with-deps chromium \
+    && crawl4ai-doctor \
     && mkdir -p /home/appuser/.cache/ms-playwright \
     && cp -r /root/.cache/ms-playwright/chromium-* \
         /root/.cache/ms-playwright/chromium_headless_shell-* \
         /home/appuser/.cache/ms-playwright/ \
     && chown -R appuser:appuser /home/appuser/.cache/ms-playwright \
-    && rm -rf /root/.cache/ms-playwright
-
-RUN crawl4ai-doctor \
+    && rm -rf /root/.cache/ms-playwright \
     && find "${APP_HOME}" -maxdepth 1 -type f -name '*.core' -delete
 
 # Ensure all cache directories belong to appuser
--- a/deploy/docker/tests/test_security_container_posture.py
+++ b/deploy/docker/tests/test_security_container_posture.py
@@ -46,8 +46,9 @@ class TestDockerfile:
             stripped = line.strip()
             if stripped.startswith("#"):
                 continue
-            assert not re.match(r"EXPOSE\s+.*\b6379\b", stripped), \
-                "redis port 6379 must not be EXPOSEd"
+            assert not re.match(
+                r"EXPOSE\s+.*\b6379\b", stripped
+            ), "redis port 6379 must not be EXPOSEd"
 
     def test_app_dir_root_owned_readonly(self, dockerfile):
         assert "chown -R root:root ${APP_HOME}" in dockerfile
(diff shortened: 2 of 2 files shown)


nightcityblade:
CI follow-up: both Security jobs pass. The only failure is the Discord notification job, whose log reports that DISCORD_WEBHOOK is not defined; this is a repository workflow-secret issue rather than a failure in the Dockerfile change.

ntohidi:
Thanks for turning this around so fast. The core-dump part is right — moving that cleanup into the `crawl4ai-doctor` layer is exactly what's needed.

Two problems though, and they're both about layer placement.

The `rm -rf /root/.cache/ms-playwright` is one layer too late. Docker only gives the space back if you delete in the same layer that made the file, and the browsers don't come from `playwright install --with-deps` — they come from `RUN crawl4ai-setup` on the line above, which calls `playwright install --force chromium` internally. I measured it on `develop`: that setup layer is 1.26GB. So after this change `du` inside the container looks clean, but `docker images` stays about where it was.

The other one is subtler. `crawl4ai-doctor` runs as root and needs `/root/.cache/ms-playwright`, which the previous layer just deleted. The build still passes, which is what makes it easy to miss — `run_doctor()` catches everything and `doctor()` exits 0 no matter what (`crawl4ai/install.py:167-211`). So it quietly stops checking anything.

While you're in there, one more: `playwright install` without naming a browser also pulls Firefox and WebKit. Neither gets copied to `appuser`, so they just sit in the image — 909MB on arm64.

Putting it all in one layer, with the doctor running before the delete:

```dockerfile
RUN crawl4ai-setup \
    && playwright install --with-deps chromium \
    && crawl4ai-doctor \
    && mkdir -p /home/appuser/.cache/ms-playwright \
    && cp -r /root/.cache/ms-playwright/chromium-* \
        /root/.cache/ms-playwright/chromium_headless_shell-* \
        /home/appuser/.cache/ms-playwright/ \
    && chown -R appuser:appuser /home/appuser/.cache/ms-playwright \
    && rm -rf /root/.cache/ms-playwright \
    && find "${APP_HOME}" -maxdepth 1 -type f -name '*.core' -delete
```

I built that on arm64 and got 9.03GB → 6.46GB. Checked the result too: root cache gone, no Firefox or WebKit, doctor passes during the build, and crawls still work as `appuser`.

Last thing — the new test asserts the cleanup lives in the `playwright install` layer, which would lock in the wrong spot. Asserting that the install, the copy, the doctor and both cleanups all share one `RUN` would catch the real thing.

nightcityblade:
Thanks for the detailed layer analysis — addressed in 14291fa.

- `crawl4ai-setup`, Chromium-only dependency installation, the doctor, runtime-cache copy, and both cleanups now run in one `RUN`.
- `crawl4ai-doctor` runs before the root Playwright cache is deleted.
- The posture test now requires all of those steps to share the setup layer and preserves their required order.

Validation:

- `python -m pytest deploy/docker/tests/test_security_*.py -q` — 317 passed, 1 xfailed
- `python -m black --check --target-version py312 deploy/docker/tests/test_security_container_posture.py` — passed
- `python -m ruff check deploy/docker/tests/test_security_container_posture.py` — passed
- `git diff --check` — passed

I could not rerun the full image build on this machine because its Docker daemon is unavailable.

Issue #1142, opened 2025-05-21
[Bug]: TypeError: 'NoneType' object has no attribute 'browser_type' when creating a new browser profile


yaniktfl:
### crawl4ai version

0.6.0

### Expected Behavior

The browser should open, allowing me to set up the profile by logging into websites and configuring settings.

### Current Behavior

When running the identity_based_browsing.py example and attempting to create a new browser profile in interactive mode, the application crashes with a TypeError indicating that a 'NoneType' object has no attribute 'browser_type', after entering the profile name
The application crashes with the following error:
[DEMO].... ℹ Error in example: 'NoneType' object has no attribute 'browser_type'

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
1. Activate virtual environment
2. Run python docs/examples/identity_based_browsing.py
3. Choose interactive mode by entering i
4. Select option 1 to create a new profile
5. Enter a name for the new profile
```

### Code snippets

```python

```

### OS

macOS

### Python version

3.12

### Browser

Chrome

### Error logs & Screenshots (if applicable)

Instructions:
[PROFILE]. ℹ 1. A browser window will open for you to set up your profile.
[PROFILE]. ℹ 2. Log in to websites, configure settings, etc. as needed.
[PROFILE]. ℹ 3. When you're done, press 'q' in this terminal to close the browser.
[PROFILE]. ℹ 4. The profile will be saved and ready to use with Crawl4AI.
[PROFILE]. ℹ ================================================================================

[DEMO].... ℹ Error in example: 'NoneType' object has no attribute 'browser_type'

Issue #1139, opened 2025-05-20
[Bug]: error while starting browser service


github-gamma:
### crawl4ai version

0.6.3

### Expected Behavior

> crwl browser start 
╭─────────────────────────────────────────────────────────────────────────────────────────────────────── Builtin Browser Start ────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Starting builtin browser                                                                                                                                                                                                             │
│                                                                                                                                                                                                                                      │
│ Browser type: chromium                                                                                                                                                                                                               │
│ Debugging port: 9222                                                                                                                                                                                                                 │
│ Headless: Yes                                                                                                                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
Error starting builtin browser: 'NoneType' object has no attribute 'browser_type'

### Current Behavior

Error starting builtin browser: 'NoneType' object has no attribute 'browser_type'

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python

```

### OS

windows

### Python version

3.12

### Browser

_No response_

### Browser version

_No response_

### Error logs & Screenshots (if applicable)

_No response_

github-gamma:
![Image](https://github.com/user-attachments/assets/42596d5b-7cbb-4fdb-b88b-f87bcfdbf24f)

![Image](https://github.com/user-attachments/assets/df9d9bef-d103-40eb-b20d-03875a082a38)

github-gamma:
![Image](https://github.com/user-attachments/assets/f2cfdc75-84cd-49f9-acb9-f7b1613bcffd)

github-gamma:
![Image](https://github.com/user-attachments/assets/2a315ab5-4be6-4bc5-a65f-5baed0ac8fa7)

ntohidi:
There is another issue related to this, but using the library itself, not the CLI: https://github.com/unclecode/crawl4ai/issues/1142

ntohidi:
When a user tries to start the browser using the CLI command, the following error is raised:

> Error starting built-in browser: 'NoneType' object has no attribute 'browser_type'

Example usage:

```
> crwl browser start
────────────────── Builtin Browser Start ────────────────── 
Starting builtin browser

Browser type: chromium
Debugging port: 9222
Headless: Yes
[...]

PR #2104, opened 2026-07-27
Fix builtin browser failing to start with 'NoneType' object has no attribute 'browser_type'

Liang-HZ:
### Summary

`crwl browser start` and `BrowserProfiler.launch_builtin_browser()` fail unconditionally with:

```
Error starting builtin browser: 'NoneType' object has no attribute 'browser_type'
```

`ManagedBrowser.__init__` still declares — and documents — the legacy keyword
arguments (`browser_type`, `user_data_dir`, `headless`, `host`, `debugging_port`,
`cdp_url`), but its body reads every value from `browser_config`, which defaults
to `None`:

```python
def __init__(self, browser_type="chromium", user_data_dir=None, headless=False,
             logger=None, host="localhost", debugging_port=9222,
             cdp_url=None, browser_config: Optional[BrowserConfig] = None):
    ...
    self.browser_type = browser_config.browser_type   # AttributeError when None
```

`launch_builtin_browser()` is the one caller that still passes only the legacy
kwargs, so it raises every time. The two other call sites in the same file were
already migrated to `browser_config=` (the old kwargs are still sitting there
commented out), which suggests this one was simply missed during the refactor.

### Fix

Two small changes:

1. `browser_profiler.py` — `launch_builtin_browser()` now builds a `BrowserConfig`
   and passes it, matching the two sibling call sites in the same file.
2. `browser_manager.py` — `ManagedBrowser.__init__` falls back to the documented
   legacy keyword arguments when `browser_config is None`, instead of raising an
   opaque `AttributeError`. This keeps the published signature honest and protects
   any third-party code still calling it the old way.

### Reproduction

Before:

```console
$ crwl browser start
Error starting builtin browser: 'NoneType' object has no attribute 'browser_type'
```

After:

```console
$ crwl browser start
[BUILTIN]. ℹ Builtin browser launched at CDP URL: http://localhost:9222
╭─────────────────────── Builtin Browser Start ───────────────────────╮
│ Builtin browser started successfully                                │
│ CDP URL: http://localhost:9222                                      │
╰─────────────────────────────────────────────────────────────────────╯
```

`crwl browser status` / `crwl browser stop` and `BrowserConfig(browser_mode="builtin")`
work against the started instance afterwards.

Also verified that constructing `ManagedBrowser(...)` with only the legacy kwargs
no longer raises, and that passing `browser_config=` behaves exactly as before.

### Related issues

- #1139 — "error while starting browser service", same error message, reported
  against 0.6.3 via `crwl browser start`. Closed, but the failure still reproduces
  on `main`.
- #1142 — closed as a duplicate of #1139.

Verified against `main` at the time of writing: both the missing `browser_config`
in `launch_builtin_browser()` and the unconditional `browser_config.*` reads in
`ManagedBrowser.__init__` are still present.

### DIFF
--- a/crawl4ai/browser_manager.py
+++ b/crawl4ai/browser_manager.py
@@ -159,6 +159,19 @@ class ManagedBrowser:
             cdp_url (str or None): CDP URL to connect to the browser. Default: None.
             browser_config (BrowserConfig): Configuration object containing all browser settings. Default: None.
         """
+        if browser_config is None:
+            # The legacy keyword arguments above are still part of this
+            # signature and are documented in the docstring; honour them
+            # instead of raising an opaque AttributeError on ``None``.
+            browser_config = BrowserConfig(
+                browser_type=browser_type,
+                user_data_dir=user_data_dir,
+                headless=headless,
+                host=host,
+                debugging_port=debugging_port,
+                cdp_url=cdp_url,
+            )
+
         self.browser_type = browser_config.browser_type
         self.user_data_dir = browser_config.user_data_dir
         self.headless = browser_config.headless
--- a/crawl4ai/browser_profiler.py
+++ b/crawl4ai/browser_profiler.py
@@ -1202,12 +1202,15 @@ class BrowserProfiler:
         os.makedirs(user_data_dir, exist_ok=True)
         
         # Create managed browser instance
-        managed_browser = ManagedBrowser(
+        browser_config = BrowserConfig(
             browser_type=browser_type,
             user_data_dir=user_data_dir,
             headless=headless,
+            debugging_port=debugging_port,
+        )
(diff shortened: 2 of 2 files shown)


eeshsaxena (commented):
Good, this closes the None path from both sides. `ManagedBrowser.__init__` now builds a `BrowserConfig` from the individual params when `browser_config is None`, so the legacy `browser_type=...` call style no longer lands on `None.browser_type`, and `launch_builtin_browser` is updated to pass a real `browser_config` instead of the individual kwargs it used before, which is what was leaving `browser_config=None` and triggering the crash. The fallback captures all six init params (browser_type, user_data_dir, headless, host, debugging_port, cdp_url), so nothing an individual-param caller passed gets dropped. Reads correct.

PR #2107, opened 2026-07-30
fix: restore upstream LiteLLM dependency

### LINKED ISSUE #2098: unclecode-litellm shadows the upstream litellm package; the quarantine that motivated the fork has been lifted
Follow-up to #1864, where the `unclecode-litellm` fork was adopted. The emergency that
motivated it has passed, and the fork now costs downstream users something it did not
cost back then.

## The fork's premise no longer holds

When #1864 was filed, PyPI had quarantined the **entire** `litellm` package — every
version was uninstallable, so pinning a fork was the only way to keep `pip install
crawl4ai` working.

That is no longer the case. On PyPI today:

- `litellm` latest is **1.93.0**
- `1.82.7`, `1.82.8`, `1.82.9` are gone — exactly the compromised releases
- `1.82.6` and everything from `1.83.0` onward are present and installable

The quarantine was lifted and only the bad releases were removed.

## What the fork costs now

`crawl4ai` 0.9.2 requires `unclecode-litellm==1.81.13`, which installs a top-level
`litellm` package — the same import name shipped by the upstream distribution. An
environment that needs both ends up with two different versions written into one
`site-packages/litellm/`, and which files survive is decided by install order rather
than by either project.

Nothing errors at install time. It surfaces later as:

```
ImportError: cannot import name '_redact_string' from 'litellm._logging'
```

The same dependency pair passed on one CI run and failed on another two minutes later,
with nothing changed but resolution order. Because the two distributions have different
names, pip and uv see no conflict to report — they install both, and one silently
overwrites the other's files.

This is not fixable downstream. A `[tool.uv] override-dependencies` entry works for a
source install, but overrides are not recorded in wheel metadata, so anyone installing
a dependent package from PyPI resolves crawl4ai's requirements directly and hits the
collision again.

## Suggested fix

Return to upstream `litellm` with a floor that excludes the compromised window:

```toml
litellm>=1.83.0
```

`1.83.0` is the first release after the three removed versions, so the floor cannot
resolve to a compromised build even if one were ever restored. This keeps `pip install
crawl4ai` exactly as it is today — full package, LLM support included, no extras to
remember — while removing the shadowing.

If there is a reason to stay on the fork, the collision can also be resolved by having
it install as `unclecode_litellm` and importing it under an alias:

```python
import unclecode_litellm as litellm
```

Same outcome for crawl4ai users, and the fork stops occupying a name it doesn't own.

A third option, if the fork is meant as a global drop-in replacement: publish the
crawler half as its own distribution (`crawl4ai-core` or similar), leaving `crawl4ai`
as the full LLM-enabled package most people install.

Happy to send a PR for whichever direction you prefer.

nightcityblade:
## Summary

Fixes #2098.

Replace the `unclecode-litellm` fork with the upstream `litellm` distribution. The lower bound starts after the removed compromised releases, and the temporary `<1.92.0` bound keeps installs on the last universal wheel release; LiteLLM 1.92+ does not currently publish macOS wheels and otherwise requires a local Rust build.

- `pyproject.toml` - Declare the safe upstream LiteLLM range in package metadata.
- `requirements.txt` - Keep development/install requirements synchronized with package metadata.

## How Has This Been Tested?

- `uvx --with 'packaging>=24.2' --from validate-pyproject validate-pyproject pyproject.toml`
- `uv pip compile pyproject.toml --no-deps --quiet` (resolves `litellm==1.91.4`)
- `uv run --python 3.12 --with toml python tests/check_dependencies.py`
- Imported both `crawl4ai` and upstream `litellm` from the resolved project environment.
- Built the wheel and verified its `Requires-Dist` metadata contains only `litellm<1.92.0,>=1.83.0`.

## Checklist:

- [ ] My code is commented, particularly in hard-to-understand areas — N/A: no executable code changed
- [ ] I have made corresponding changes to the documentation — N/A: the dependency declarations are the affected configuration
- [ ] I have added/updated unit tests that prove my fix is effective or that my feature works — N/A: this is package metadata; wheel metadata and a resolved import were tested directly
- [x] New and existing relevant tests pass locally with my changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "aiosqlite~=0.20",
     "anyio>=4.0.0",
     "lxml>=5.3,<7",
-    "unclecode-litellm==1.81.13",
+    "litellm>=1.83.0,<1.92.0",
     "numpy>=1.26.0,<3",
     "pillow>=10.4",
     "playwright>=1.49.0",
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ aiohttp>=3.11.11
 aiosqlite~=0.20
 anyio>=4.0.0
 lxml>=5.3,<7
-unclecode-litellm==1.81.13
+litellm>=1.83.0,<1.92.0
 numpy>=1.26.0,<3
 pillow>=10.4
 playwright>=1.49.0


nightcityblade:
CI note: the only failing check is the Discord notification job. Its log shows `Secret source: None`, an empty `DISCORD_WEBHOOK`, and then `Env var DISCORD_WEBHOOK is not defined`; it does not install or test this PR’s dependency changes. This is a fork-secret/workflow configuration failure rather than a code failure in the patch.

PR #2109, opened 2026-07-30
fix: only treat the base domain and its subdomains as internal

eeshsaxena:
## Summary

`is_external_url` decided whether a link belongs to the site being crawled with a bare suffix check:

```python
return not url_domain.endswith(base)
```

Any host whose name merely *ends with* the base domain string passed that check and was classified as **internal**:

```python
is_external_url("http://malicious-example.com", "example.com")  # False (treated as internal)
is_external_url("http://notexample.com/path", "example.com")    # False (treated as internal)
```

`base` comes from `get_base_domain()`, so it is a bare registrable domain and these are genuinely different sites an attacker can register. The result feeds link classification in `content_scraping_strategy.py` and the deep-crawl scope check in `domain_mapper.py`, so a crawl scoped to one domain could follow links onto look-alike domains.

The fix matches what the rest of the codebase already does (`deep_crawling/filters.py`, `domain_mapper.py`): **internal means equal to the base domain, or a dot-separated subdomain of it** (`domain == base or domain.endswith(f".{base}")`). Real subdomains (`blog.example.com`) and relative URLs are unchanged. The `www.` strip also moves from `.replace("www.", "")` to `.removeprefix("www.")`, so a host containing `www.` somewhere other than the front is not silently rewritten.

No existing issue filed — happy to open one first if you'd prefer it tracked separately.

- `crawl4ai/utils.py` — rewrite `is_external_url`'s membership test from `endswith(base)` to `== base or endswith("." + base)`, and switch the `www.` strip to `removeprefix`.
- `tests/regression/test_reg_utils.py` — add `TestIsExternalUrl` covering the base domain, a subdomain, `www.`, the two look-alike cases (`notexample.com`, `malicious-example.com`), the base domain as a prefix, an unrelated domain, a relative URL, and a `mailto:` link.

## How Has This Been Tested?

Ran `pytest tests/regression/test_reg_utils.py`. Against the current code, the two look-alike cases in the new `TestIsExternalUrl` fail (demonstrating the bug); with the fix applied the full file passes.

> Note: `tests/test_link_extractor.py::test_link_extractor` fails on a clean checkout as well (the async test isn't picked up by the configured asyncio plugin), so it's unrelated to this change.

## Checklist:

### DIFF
--- a/crawl4ai/utils.py
+++ b/crawl4ai/utils.py
@@ -2552,12 +2552,18 @@ def is_external_url(url: str, base_domain: str) -> bool:
         if not parsed.netloc:  # Relative URL
             return False
 
-        # Strip port and 'www.' from both domains for comparison
-        url_domain = parsed.netloc.lower().split(":")[0].replace("www.", "")
-        base = base_domain.lower().split(":")[0].replace("www.", "")
+        # Strip port and a leading 'www.' from both domains for comparison.
+        # removeprefix (not replace) so a host containing 'www.' somewhere
+        # other than the front is left untouched.
+        url_domain = parsed.netloc.lower().split(":")[0].removeprefix("www.")
+        base = base_domain.lower().split(":")[0].removeprefix("www.")
 
-        # Check if URL domain ends with base domain
-        return not url_domain.endswith(base)
+        # Internal means the base domain itself or one of its subdomains. A
+        # bare endswith(base) also matches look-alike domains that merely end
+        # with the base string (notexample.com for example.com), which are
+        # genuinely different sites an attacker can register.
+        is_internal = url_domain == base or url_domain.endswith(f".{base}")
+        return not is_internal
     except Exception:
         return False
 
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
This closes a real hole. The old `url_domain.endswith(base)` treated any look-alike suffix as internal, so `malicious-example.com` and `example.com.evil.com` both counted as internal to `example.com`. Switching to `url_domain == base or url_domain.endswith(f".{base}")` requires either an exact match or a dot-boundary subdomain, which is the correct rule, and moving from `.replace("www.", "")` (which strips `www.` anywhere) to `.removeprefix("www.")` avoids mangling hosts that merely contain `www`. The test matrix covers the important cases, including the `example.com.evil.com` prefix trick.

One robustness nit, not a blocker: `parsed.netloc.lower().split(":")[0]` keeps userinfo, so `http://user@evil.com` leaves `user@evil.com` as the compared host. `parsed.hostname` would strip userinfo and port and lowercase for you, sidestepping that whole class of edge case.

eeshsaxena:
Closing this to clear out my older open PRs, since it has been sitting a while without review. No hard feelings if it was not a priority. If it is still useful to you, say the word and I will reopen it and rebase it on current main.

PR #1904, opened 2026-04-07
feat: add preserve_classes/preserve_tags whitelist to PruningContentFilter (#1900)

hafezparast:
## Summary
Adds `preserve_classes` and `preserve_tags` params to `PruningContentFilter` so users can protect specific elements (author names, timestamps, attribution) from being pruned.

Addresses #1900

## The Problem
`md-fit` uses density-based scoring to strip boilerplate. Short metadata elements (usernames, bylines, timestamps) score low and get removed alongside actual boilerplate — losing "who said what" on discussion pages.

## The Fix
Two new optional params on `PruningContentFilter`:

```python
PruningContentFilter(
    preserve_classes=["author", "byline", "comment-header"],
    preserve_tags=["time", "cite"],
)
```

Whitelisted nodes skip scoring entirely — always kept. Default is empty sets, so existing behavior is unchanged.

## Usage
```python
from crawl4ai import AsyncWebCrawler, CrawlerRunConfig, DefaultMarkdownGenerator
from crawl4ai.content_filter_strategy import PruningContentFilter

config = CrawlerRunConfig(
    markdown_generator=DefaultMarkdownGenerator(
        content_filter=PruningContentFilter(
            preserve_classes=["author", "byline", "username"],
            preserve_tags=["time", "cite"],
        )
    )
)

async with AsyncWebCrawler() as crawler:
    result = await crawler.arun(url="https://gist.github.com/...", config=config)
    print(result.markdown.fit_markdown)  # Now includes author names
```

## Test plan
- [x] 20 unit tests passing
- [x] Existing pipeline tests pass (16/16)
- [x] Default behavior unchanged (empty whitelist)
- [x] Nav/footer still removed (excluded_tags runs before pruning)
- [x] Nonexistent classes/tags in whitelist are harmless

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1900, opened 2026-04-06
[Bug]: md-fit strips meaningful content metadata (usernames, attribution) not just boilerplate


mattheworiordan:
### crawl4ai version

0.8.6

### Expected Behavior

`md-fit` should strip page chrome (navigation, footers, sidebars, cookie banners) but preserve meaningful content metadata like author names and attribution on user-generated content. Comment attribution (who said what) is semantic content, not boilerplate.

### Current Behavior

`md-fit` strips comment author usernames and profile links from GitHub pages while keeping the comment body and date. The output shows `commented Apr 6, 2026` with no indication of who wrote it.

With `-o md`, you get:

```
**[mattheworiordan](https://gist.github.com/mattheworiordan)** commented Apr 6, 2026
```

With `-o md-fit`, you get:

```
commented Apr 6, 2026
```

The username and profile link are gone. On pages with multiple commenters, it's impossible to tell who said what.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- URL: https://gist.github.com/mattheworiordan/99ba717f7a8ca5a7f838913722cfe7ac
- Settings: `-o md-fit` (default fit mode, no other flags)
```

### Steps to Reproduce

```bash
# 1. Crawl the gist with full markdown - username preserved
crwl crawl https://gist.github.com/mattheworiordan/99ba717f7a8ca5a7f838913722cfe7ac -o md

# Look for: **[mattheworiordan](...)** commented Apr 6, 2026

# 2. Same URL with md-fit - username stripped
crwl crawl https://gist.github.com/mattheworiordan/99ba717f7a8ca5a7f838913722cfe7ac -o md-fit

# Look for: commented Apr 6, 2026  (no username)
```

### Code snippets

```python
# Not a code bug - this is CLI / markdown generation behaviour.
# Reproduces purely via crwl CLI as shown above.
```

### OS

macOS (Darwin 25.2.0)

### Python version

3.14.3

### Browser

(default Chromium managed by crawl4ai)

### Browser version

(crawl4ai default)

### Error logs & Screenshots (if applicable)

n/a

hafezparast:
This is expected behavior of the `PruningContentFilter` — it scores each HTML block on text density, link ratio, tag weight, and class patterns. Short metadata elements (usernames, timestamps) score low on density and get pruned alongside actual boilerplate.

[...]

Issue #1181, opened 2025-06-04
[Bug]: Incorrect crawlered code format of `import xxx`


haoyang9804:
### crawl4ai version

0.5.0.post2

### Expected Behavior

When crawling code blocks from the triton tutorial page: https://triton-lang.org/main/getting-started/tutorials/01-vector-add.html#sphx-glr-getting-started-tutorials-01-vector-add-py, space between `import` and the package name is omitted.

The webpage contains several `import` statement:

```
import torch

import triton
import triton.language as tl
```

The crawlered results should contain exactly same code snippet.

### Current Behavior

The crawlered `import`-related results are

```
importtorch
importtriton
importtriton.languageastl
```

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
A simple reproducible script:


import asyncio
from crawl4ai import *

async def main():
    # Create an instance of AsyncWebCrawler
    async with AsyncWebCrawler() as crawler:
        # Run the crawler on a URL
        result = await crawler.arun(url="https://triton-lang.org/main/getting-started/tutorials/01-vector-add.html#sphx-glr-getting-started-tutorials-01-vector-add-py")

        # Print the extracted content
        print(result.markdown)

asyncio.run(main())
```

### Steps to Reproduce

```bash
run the script, and you can find the incorrect `import` results.
```

### Code snippets

```python

```

### OS

macOS

### Python version

3.11.11

### Browser

Chrome

ntohidi:
## Root Cause

In `crawl4ai/content_scraping_strategy.py:763`, the scraping strategy calls:

`self.remove_empty_elements_fast(body, 1)`

This method removes elements with text content below the word threshold.
Syntax-highlighted code blocks use <span> tags for each token, including whitespace:

 `<span class="kn">import</span><span class="w"> </span><span class="nn">torch</span>`

The whitespace span `<span class="w"> </span>`:
- Has text content of `" "` which becomes `""` after `.strip()`
- Contains 0 words (< threshold of 1)
- Has no child elements

Therefore, it was being removed, losing the space between tokens.

ntohidi:
fixed in the `develop` branch. will be in the next release in 2 weeks.

PR #2111, opened 2026-07-30
fix: skip pruning inside <pre>/<code> in PruningContentFilter

### LINKED ISSUE #2110: PruningContentFilter drops whitespace-only spans in <pre>/<code> — #1181's bug, still present on the fit_markdown path
## Summary

#1181 ("importtorch") was fixed in v0.7.8 by skipping elements inside `<pre>`/`<code>` in `remove_empty_elements_fast` (content_scraping_strategy.py). `PruningContentFilter` has the same failure, unfixed: it scores and decomposes whitespace-only token spans inside highlighted code, so `fit_markdown` corrupts code that `raw_markdown` (post-0.7.8) preserves.

## Environment

crawl4ai 0.9.2, Docker image `unclecode/crawl4ai:0.9.2`, REST `/crawl`. Code references are v0.9.2 (unchanged on `develop`).

## Repro

`crawler_config` → `markdown_generator: DefaultMarkdownGenerator` with `content_filter: PruningContentFilter(threshold=0.48, threshold_type="fixed")`. Fetch any page with `<span>`-based syntax highlighting and compare `fit_markdown` vs `raw_markdown`:

- https://kubernetes.io/docs/concepts/services-networking/service/ — every YAML block collapses to one line: `apiVersion:v1kind:Servicemetadata:name:my-service…`
- https://dev.to/godofgeeks/multi-stage-builds-in-docker-g4i — `FROM golang:1.21 AS builder` → `FROMgolang:1.21ASbuilder`; `&&` dropped.
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce — all code blocks removed entirely (whole `<pre>` subtrees pruned).

`raw_markdown` is intact in the same responses.

## Root cause

`content_filter_strategy.py`: `_prune_tree` recurses into kept nodes and `decompose()`s any child scoring under threshold. For a highlighter whitespace span (`<span class="w"> </span>`), `_compute_composite_score` yields: text_density 0 (text_len 0) · link_density term 0.20 · tag_weight `span`=0.3 → 0.06 · text_length `log(1)`=0 → composite ≈ 0.26–0.36 < 0.48 → removed. This is the same arithmetic the v0.7.8 scraper fix bypasses via its `<pre>`/`<code>` guard ("This preserves whitespace-only spans (e.g., `<span class="w"> </span>`) in code blocks"); the filter never got the guard. Low-scoring whole `<pre>` blocks (huge tag_len from hundreds of spans → near-zero text density) are likewise decomposed (the MDN case).

## Suggested fix

Port the scraper's guard into `_prune_tree`: skip scoring/pruning for `<pre>`/`<code>` subtrees. PR incoming.

## Workaround (v0.9.1+)

`PruningContentFilter(..., preserve_tags=["pre","code"])` (added by #1904) protects the whole subtree; or clients can fall back to `raw_markdown` when `fit_markdown` loses code fences.

kinorai:
Fixes #2110

## What

`PruningContentFilter._prune_tree` now skips `<pre>`/`<code>` subtrees entirely — never scored, never pruned. This mirrors the scraper-side guard that #1181 added to `WebScrapingStrategy.remove_empty_elements_fast` in v0.7.8, which fixed `raw_markdown` but left `fit_markdown` unprotected.

## Why

Syntax highlighters wrap every token in `<span>`, including whitespace-only spans (`<span class="w"> </span>`). Those score ~0.26–0.36 against the default 0.48 threshold (text_density 0, tag_weight span=0.3, log(1)=0) and get decomposed — `import torch` becomes `importtorch`, `FROM golang:1.21 AS builder` becomes `FROMgolang:1.21ASbuilder` in `fit_markdown`. Heavily-highlighted `<pre>` blocks (hundreds of spans → huge tag_len, near-zero text density) can be dropped wholesale, which removes all code from e.g. MDN pages. Details and real-URL repros in #2110.

## Tests

`tests/test_pruning_code_whitespace.py` (4 tests, modeled on `test_pruning_preserve_whitelist_1900.py`):
- whitespace-only spans inside code survive filtering
- a heavily-highlighted `<pre>` block is not dropped
- normal pruning outside code blocks still works (nav is still pruned)
- rendered text keeps token spacing (`import torch`, `FROM golang:1.21 AS builder`)

Without the fix, the two whitespace tests fail with the exact corruption signature; with it, all 24 pruning-filter tests pass (including the full #1900 preserve-whitelist suite — no regressions).

`preserve_tags=["pre","code"]` (#1904) remains a valid explicit workaround for older 0.9.x versions; this PR makes the safe behavior the default.

### DIFF
--- a/crawl4ai/content_filter_strategy.py
+++ b/crawl4ai/content_filter_strategy.py
@@ -712,6 +712,14 @@ class PruningContentFilter(RelevantContentFilter):
         if self._is_preserved(node):
             return
 
+        # Skip pruning inside <pre>/<code> blocks where whitespace is significant.
+        # Syntax highlighters wrap every token in <span> (including whitespace-only
+        # spans like <span class="w"> </span>) which score far below any threshold
+        # and get decomposed, corrupting code. Mirrors the scraper-side guard in
+        # WebScrapingStrategy.remove_empty_elements_fast (#1181).
+        if node.name in ("pre", "code"):
+            return
+
         text_len = len(node.get_text(strip=True))
         tag_len = len(node.encode_contents().decode("utf-8"))
         link_text_len = sum(
--- a/tests/test_pruning_code_whitespace.py
+++ b/tests/test_pruning_code_whitespace.py
@@ -1,0 +1,78 @@
+"""
+PruningContentFilter must not prune inside <pre>/<code> blocks.
+
+Syntax highlighters (pygments, prism, rouge) wrap every token in <span>,
+including whitespace-only spans like <span class="w"> </span>. Those spans
+score ~0.26-0.36 against the default 0.48 threshold and were decomposed,
+collapsing `import torch` to `importtorch` — the fit_markdown counterpart of
+the scraper-side bug fixed in #1181. Heavily-highlighted <pre> blocks (huge
+tag_len, low text density) could also be dropped wholesale.
+"""
+import pytest
+from crawl4ai.content_filter_strategy import PruningContentFilter
+
+
+# Pygments-style highlighting: every token in a span, whitespace in class="w".
+HIGHLIGHTED_CODE_HTML = """
+<html><body>
+<article>
+    <h1>Using PyTorch</h1>
+    <p>This paragraph carries enough real prose to pass the pruning threshold
+    comfortably. It explains how to import the library and configure a device
+    before running the model, so the surrounding article content is retained
+    by the filter as fully relevant text.</p>
+    <pre><code class="highlight"><span class="kn">import</span><span class="w"> </span><span class="nn">torch</span>
+<span class="kn">import</span><span class="w"> </span><span class="nn">triton.language</span><span class="w"> </span><span class="k">as</span><span class="w"> </span><span class="nn">tl</span>
+<span class="n">FROM</span><span class="w"> </span><span class="n">golang</span><span class="p">:</span><span class="mf">1.21</span><span class="w"> </span><span class="n">AS</span><span class="w"> </span><span class="n">builder</span></code></pre>
+    <p>Another paragraph of real prose after the code block, long enough to be
+    kept on its own merits, describing what the snippet above actually does in
+    the larger training pipeline.</p>
+</article>
+<nav>
+    <a href="/">Home</a>
+    <a href="/docs">Docs</a>
+</nav>
+</body></html>
+"""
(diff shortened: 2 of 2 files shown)


eeshsaxena (commented):
Right fix for the highlighter-span problem. Returning early when `node.name in ("pre", "code")` skips pruning the block and, since `_prune_tree` recurses into children, its whole subtree too, so the per-token `<span class="w"> </span>` whitespace spans that score below the 0.48 threshold can no longer be decomposed into `importtorch`. Placing it after the `_is_preserved` check keeps the existing whitelist behavior intact.

The one tradeoff is that an oversized or genuinely noisy `<pre>` (huge tag_len, low text density) is now always kept rather than droppable, but for code fidelity that is the right call, and it matches the scraper-side fix in #1181.

erekola:
I found an ancestor-pruning case at `f8181ae`: the guard preserves a short highlighted `x=1` block directly inside the article, but wrapping that same block in `<div class="highlight">` still removes it. The article survives in both cases.

The following fixture uses the default filter settings:

```python
from bs4 import BeautifulSoup
from crawl4ai.content_filter_strategy import PruningContentFilter

prose = "<p>" + "This article explains how to import libraries and run examples. " * 8 + "</p>"
code = '<pre><code><span class="n">x</span><span class="o">=</span><span class="mi">1</span></code></pre>'
for snippet in (code, '<div class="highlight">' + code + '</div>'):
    html = '<body><article>' + prose + snippet + prose + '</article></body>'
    out = '\n'.join(PruningContentFilter().filter_content(html))
    pre = BeautifulSoup(out, 'lxml').find('pre')
    print(pre.get_text() if pre else None)
# PR: x=1, then None
# develop at 75bacbd: None, then None
```

The wrapper is scored and decomposed before recursion reaches the new guard. A wrapped-code fixture would make this boundary explicit alongside the whitespace tests. This also limits the existing `preserve_tags=["pre", "code"]` workaround when an ancestor is pruned.

I ran an isolated source-level probe using the unmodified `RelevantContentFilter` and `PruningContentFilter` class ASTs from both checkouts with BeautifulSoup/lxml. I did not run the package initializer, crawler or full test suite. Longer `import torch` fixtures confirmed the whitespace fix in direct and wrapped blocks. AI assisted this check.

PR #2117, opened 2026-08-01
fix(docker): preserve single-url crawl failure details

### LINKED ISSUE #2116: [Bug]: `/md` and `/llm/{url}` return HTTP 500 when anti-bot detection marks a result failed
**Version:** 0.9.2 (Docker), also present on `main` @ `7e80152`

## What happens

`handle_markdown_request` in `deploy/docker/api.py` promotes any unsuccessful crawl into a server error:

```python
if not result.success:
    raise HTTPException(
        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
        detail=result.error_message
    )
```

`handle_llm_qa` carries the identical block.

Since 0.8.5, `is_blocked()` sets `success = False` and `error_message = "Blocked by anti-bot protection: <reason>"` whenever the escalation chain cannot rescue a page. So a remote site serving a Cloudflare, DataDome or Akamai challenge — an ordinary and fully expected outcome — surfaces from the Docker API as a 500.

## Why the caller can't tell

`_http_exception_handler` in `deploy/docker/server.py` genericizes every 5xx to avoid leaking internals:

```json
{"error": "Internal server error", "correlation_id": "<random>"}
```

That hardening is right, and MIGRATION.md documents it deliberately. But because a bot-block is routed through 5xx, the sanitiser strips the one field the caller needs. `error_message` is computed, logged, then discarded before it reaches the client. Two individually correct decisions collide.

The result: an HTTP client cannot distinguish "the target site blocked us" from "the service is unhealthy" without shell access to the container logs.

## Expected

A blocked page is a crawl outcome, not a server fault. The [anti-bot docs](https://docs.crawl4ai.com/advanced/anti-bot-and-fallback/) define the terminal state as `success=False` plus an `error_message` naming the block reason.

`handle_crawl_request` already honours this — it returns 200 with per-URL `success` flags and reasons intact. The single-URL endpoints disagree with both the documentation and their sibling handler in the same file.

Suggested: 502, or 200 with a structured failure body; either way preserving `error_message`.

## Repro

```bash
curl -sS -X POST http://localhost:11235/md \
  -H "Content-Type: application/json" \
  -d '{"url":"https://www.sagaftra.org/"}'
# HTTP 500
# {"error":"Internal server error","correlation_id":"e2996592a617"}

curl -sS -X POST http://localhost:11235/crawl \
  -H "Content-Type: application/json" \
  -d '{"urls":["https://www.sagaftra.org/"]}'
# HTTP 200
# results[0].success       = false
# results[0].status_code   = 403
# results[0].error_message = "Blocked by anti-bot protection: DataDome captcha"
```

Same page, same crawler, same block. One endpoint says what happened; the other says nothing.

## Why this bites harder over REST than in the library

The docs state the terminal condition as "When all attempts fail and blocking is still detected, the result is returned with `success=False` and `error_message` describing the block reason." That is the contract `/crawl` honours and `/md` breaks.

The escalation those attempts refer to is unreachable over REST. `async_configs.py:203-213` places both escape hatches in `UNTRUSTED_FORBIDDEN_FIELDS`:

- `fallback_fetch_function` — additionally typed `Optional[Callable[...]]` (line 1705), so it could not cross a JSON boundary regardless
- `proxy_config`, `proxy_rotation_strategy`, `proxy_session_id`

So on a stock Docker deployment no request body can supply either one. Detection runs, escalation cannot, and every block lands as an opaque 500. Discussion #1905 — Avada/WordPress pages flagged `Structural: no tag` with no anti-bot present, still unanswered — is this same failure seen from the other end: the reporter cannot inspect what came back, because the response body no longer contains it.

## Impact

One deployment logged 132 of these over six days across roughly fifteen distinct block reasons, including several `Structural:` classifications on sites we believe are not protected at all. Per-domain retry policy or extraction routing can't be built on the single-URL endpoints while the reason is unavailable to the caller.

## Note on the environment

These figures come from a locally derived image built on the pinned 0.9.2 digest, carrying local operational patches unrelated to the paths below. The three code paths above — the `/md` failure branch, the `/llm/{url}` failure branch and the 5xx sanitiser — are unmodified upstream, and the `/md` branch is byte-identical on current `main`.

nightcityblade:
## Summary

Fixes #2116

Treat unsuccessful crawls from the single-URL `/md` and `/llm/{url}` handlers as upstream failures (`502 Bad Gateway`) so the existing exception handler preserves `error_message` instead of replacing it with an opaque internal-server-error response. The LLM handler now also preserves deliberate `HTTPException` statuses. This changes those two failure responses from HTTP 500 to HTTP 502.

- `deploy/docker/api.py` - centralize unsuccessful crawl handling and return a detail-preserving 502 from both single-URL handlers.
- `deploy/docker/tests/test_api_crawl_failures.py` - verify the helper mapping and exercise authenticated `/md` and `/llm/{url}` requests through the real FastAPI app, with a mocked failing crawler.

## How Has This Been Tested?

- `/tmp/crawl4ai-followup-venv/bin/python -m pytest deploy/docker/tests/test_api_crawl_failures.py deploy/docker/tests/test_security_headers_xss.py::TestErrorSanitization -q` — 5 passed, 4 existing warnings.
- `/tmp/crawl4ai-followup-venv/bin/ruff check --isolated --select E9,F63,F7,F82 deploy/docker/api.py deploy/docker/tests/test_api_crawl_failures.py` — passed.
- `/tmp/crawl4ai-followup-venv/bin/python -m compileall -q deploy/docker/api.py deploy/docker/tests/test_api_crawl_failures.py` — passed.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas (N/A: no complex logic was introduced)
- [x] I have made corresponding changes to the documentation (N/A: the behavioral status change is called out in this PR description)

### DIFF
--- a/deploy/docker/api.py
+++ b/deploy/docker/api.py
@@ -92,6 +92,15 @@ import psutil, time
 
 logger = logging.getLogger(__name__)
 
+
+def _raise_for_crawl_failure(result):
+    if not result.success:
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail=result.error_message,
+        )
+
+
 # --- Helper to get memory ---
 def _get_memory_mb():
     try:
@@ -147,11 +156,7 @@ async def handle_llm_qa(
         enforce_egress(browser_cfg)
         crawler = await get_crawler(browser_cfg)
         result = await crawler.arun(url)
-        if not result.success:
-            raise HTTPException(
-                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-                detail=result.error_message
-            )
+        _raise_for_crawl_failure(result)
         content = result.markdown.fit_markdown or result.markdown.raw_markdown
 
         # Create prompt and get LLM response
@@ -179,6 +184,8 @@ async def handle_llm_qa(
         )
 
         return response.choices[0].message.content
+    except HTTPException:
+        raise
     except LLMProviderNotAllowed as e:
         raise HTTPException(status_code=400, detail=str(e))
     except Exception as e:
@@ -389,11 +396,7 @@ async def handle_markdown_request(
             )
         )
 
-        if not result.success:
-            raise HTTPException(
-                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-                detail=result.error_message
-            )
+        _raise_for_crawl_failure(result)
 
         return (result.markdown.raw_markdown
                if filter_type == FilterType.RAW
--- a/deploy/docker/tests/test_api_crawl_failures.py
+++ b/deploy/docker/tests/test_api_crawl_failures.py
@@ -1,10 +1,9 @@
-import inspect
 from types import SimpleNamespace
 
 import pytest
 from fastapi import HTTPException
 
-from api import _raise_for_crawl_failure, handle_llm_qa, handle_markdown_request
+from api import _raise_for_crawl_failure
 
 
 def test_crawl_failure_is_reported_as_bad_gateway():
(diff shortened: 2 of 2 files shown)


nightcityblade:
CI note: the code-related posture-gate and security-offline checks both pass. The only failing job is the Discord notification workflow; its log shows an empty DISCORD_WEBHOOK and exits before exercising this change. This is repository/fork-secret configuration rather than a patch failure.

bong-u:
Nice work.
I hit the same 500 from a client and found one more spot the issue didn't cover.

`deploy/docker/server.py:911`:

```python
if all(not result["success"] for result in results["results"]):
    raise HTTPException(500, f"Crawl request failed: {results['results'][0]['error_message']}")
```

For a single-URL request "all failed" is just "it failed", so the detail gets logged and dropped by the same 500 branch. Stock `0.9.2`:

```
POST /crawl {"urls":["https://www.sagaftra.org/"]}
→ HTTP 500
→ {"error":"Internal server error","correlation_id":"844da2acf429"}
```

That's the same body the issue reports for `/md`, from the endpoint it lists as unaffected.
`server.py:911` is byte-identical on `v0.9.2`, `main` and this PR's head.

The detail is already interpolated here, so switching that `500` to `502` is enough to make it survive.
If you'd rather reuse `_raise_for_crawl_failure`, it won't drop in as-is: `handle_crawl_request` returns `model_dump()` dicts (`api.py:751`), so it's `result["success"]`.

nightcityblade:
Thanks for the concrete reproduction. You're right: the all-results-failed branch in /crawl also raises 500, so the issue's statement that /crawl is unaffected is too broad. This PR is scoped to the two single-URL handlers from #2116, and I don't want to expand it without maintainer direction. I'm happy to include that adjacent branch here if the maintainers prefer, or it can be tracked separately.

bong-u:
That makes sense.
My main goal was to make sure the /crawl case doesn't get overlooked.
I agree that maintainer preference should decide whether it belongs here or in a follow-up.
Thanks for confirming the behavior.

ntohidi:
@nightcityblade 
The fix looks right. 502 works because the handler at `server.py:509` only strips the body on 500s — anything else keeps its detail. And the `except HTTPException: raise` you added is needed, otherwise the `except Exception` below it turns your 502 back into a 500.

One problem with the tests. `test_single_url_handlers_use_crawl_failure_mapping` just checks that a line of text appears in the source code. That'll pass even if the endpoint is broken. So nothing here actually tests the bug — that a blocked page now comes back with its reason instead of an empty 500.

Take a look at `test_crawl_failure_response.py` in the same folder. It sends a real request and checks the response. Can you redo that test the same way? The other one's fine, leave it.

Also worth mentioning in the description that `/md` and `/llm/{url}` now return 502 instead of 500, since that affects anyone already calling them.

nightcityblade:
Thanks — addressed in `4fc16ca`. I replaced the source-inspection assertion with authenticated requests through the real FastAPI app for both `/md` and `/llm/{url}`. Each test supplies a mocked failing crawler and verifies that the client receives HTTP 502 with the original anti-bot detail.

I also updated the PR description to call out the 500-to-502 behavior change explicitly. The focused endpoint and exception-handler suite passes (5 tests).

ntohidi:
Tests look good now — thanks. Covering both endpoints and asserting the detail body is exactly right.

@bong-u on the `/crawl` branch you found: that's already fixed on `develop`. `130a382` removed those lines a few days ago. You were testing 0.9.2, where it's still there — it'll go away with the next release. Nothing to follow up on, and thanks for chasing it.

Merging this.

Issue #1367, opened 2025-08-05
[Bug]: Error: Failed on navigating ACS-GOTO: Page.goto: net::ERR_ABORTED


EstopaceMA:
### crawl4ai version

0.7.2

### Expected Behavior

It opens the browser to loading state and always getting an error.

### Current Behavior

I'm having this kind of issue and can't find a way to fix.

```
[INIT].... → Crawl4AI 0.7.2 
Loading page 1...
[ERROR]... × https://www.theknot.com...enues-atlanta-ga?page=1  | Error: Unexpected error in _crawl_web at line 763 in _crawl_web 
(../../../../opt/homebrew/Caskroom/miniconda/base/envs/deep-seek-crawler/lib/python3.12/site-packages/crawl4ai/async_crawler_strategy.py):
Error: Failed on navigating ACS-GOTO:
Page.goto: net::ERR_ABORTED at https://www.theknot.com/marketplace/wedding-reception-venues-atlanta-ga?page=1
Call log:
  - navigating to "https://www.theknot.com/marketplace/wedding-reception-venues-atlanta-ga?page=1", waiting until "domcontentloaded"


Code context:
 758                               tag="GOTO",
 759                               params={"url": url},
 760                           )
 761                           response = None
 762                       else:
 763 →                         raise RuntimeError(f"Failed on navigating ACS-GOTO:\n{str(e)}")
 764   
 765                   await self.execute_hook(
 766                       "after_goto", page, context=context, url=url, response=response, config=config
 767                   )
 768    
Error fetching page for 'No Results Found' check: Unexpected error in _crawl_web at line 763 in _crawl_web (../../../../opt/homebrew/Caskroom/miniconda/base/envs/deep-seek-crawler/lib/python3.12/site-packages/crawl4ai/async_crawler_strategy.py):
Error: Failed on navigating ACS-GOTO:
Page.goto: net::ERR_ABORTED at https://www.theknot.com/marketplace/wedding-reception-venues-atlanta-ga?page=1
Call log:
  - navigating to "https://www.theknot.com/marketplace/wedding-reception-venues-atlanta-ga?page=1", waiting until "domcontentloaded"
```

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash

```

### Code snippets

```python

```

### OS

MacOS

### Python version

3.12

### Browser

Chromium

moneebshammout:
same error, if u use proxy it always fails like that

[...]

PR #2130, opened 2026-08-07
fix(docker): add pypdf and honor stream scraping strategy

### LINKED ISSUE #2127: [Bug]: Docker image ships without pypdf, so PDFContentScrapingStrategy fails at runtime
### crawl4ai version

0.9.2

### Expected Behavior

After #1815 added `PDFContentScrapingStrategy` to `ALLOWED_DESERIALIZE_TYPES`,
`POST /crawl` with a PDF URL and that strategy should return extracted PDF content from the official Docker image.

### Current Behavior

Deserialization works now, but the request still fails. Two blockers remain that #1815 did not cover.

1. The Docker image has no `pypdf`.

`pypdf` is only a `[pdf]` extra (`pyproject.toml:62`) and is absent from `deploy/docker/requirements.txt`. With the default `INSTALL_TYPE=default` the Dockerfile runs `pip install "/tmp/project"` (Dockerfile:168), no extras, so the strategy raises at runtime:

```
ImportError: pypdf is required for PDF processing. Install with 'pip install crawl4ai[pdf]'
```
(`crawl4ai/processors/pdf/processor.py:64,80,131,239,422`)

`INSTALL_TYPE=all` pulls it in, but also torch, transformers and nltk, just to get one pure-Python dependency.

2. `/crawl/stream` silently discards the strategy.

`deploy/docker/api.py:876` overwrites whatever the client sent:

```python
crawler_config.scraping_strategy = LXMLWebScrapingStrategy()
```

The non-streaming `handle_crawl_request` path has no such line, so the same payload behaves differently on the two endpoints and the streaming one reports no error.

Related, for PDFs served as attachments: `async_crawler_strategy.py:770-778` only tolerates `net::ERR_ABORTED` when `browser_config.accept_downloads` is set. PDFs that Chromium renders inline, like arxiv's, navigate fine. Ones sent with `Content-Disposition: attachment` raise `RuntimeError: Failed on navigating ACS-GOTO` before scraping runs. See also #1367.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
`POST /crawl` against https://arxiv.org/pdf/2310.08535 with `crawler_config.params.scraping_strategy.type = "PDFContentScrapingStrategy"`.
```

### Steps to Reproduce

```bash
1. Run the official crawl4ai Docker image.
2. `POST /crawl` with the payload below, get `ImportError: pypdf is required`.
3. Send the same payload to `/crawl/stream`, get HTML-scraped empty content instead, with no error.
```

### Code snippets

```python
{
  "urls": ["https://arxiv.org/pdf/2310.08535"],
  "crawler_config": {
    "type": "CrawlerRunConfig",
    "params": {
      "cache_mode": "bypass",
      "stream": false,
      "scraping_strategy": {
        "type": "PDFContentScrapingStrategy",
        "params": {"extract_images": false, "batch_size": 8}
      }
    }
  }
}
```

### OS

Linux (Docker)

### Python version

3.12

### Browser

_No response_

### Browser version

_No response_

ImportError: pypdf is required for PDF processing. Install with 'pip install crawl4ai[pdf]

nightcityblade:
## Summary

Addresses part of #2127 without closing it.

The default Docker image now installs the PDF processor dependency, and the streaming crawl handler preserves the scraping strategy supplied by the client instead of replacing it with `LXMLWebScrapingStrategy`.

This PR does not resolve the remaining Chromium navigation failure when a PDF response triggers a download. That behavior remains tracked in #2127 (and is related to #1367).

- `deploy/docker/requirements.txt` - Install `pypdf>=6.0.0` in the default Docker image, matching the root requirement.
- `deploy/docker/api.py` - Preserve the deserialized scraping strategy for streaming crawls.
- `tests/test_issue_2127_docker_pdf.py` - Parse Docker requirement names robustly and behaviorally verify that the real stream handler passes a requested `PDFContentScrapingStrategy` to the crawler.

## How Has This Been Tested?

- `.venv/bin/pytest -q tests/test_issue_2127_docker_pdf.py tests/test_pr_1795_1798_1734.py::TestDeepCrawlStreamBranching` (7 passed)
- `.venv/bin/black --check --target-version py312 --fast tests/test_issue_2127_docker_pdf.py`
- `.venv/bin/python -m py_compile tests/test_issue_2127_docker_pdf.py deploy/docker/api.py`
- `uv pip install --dry-run --python .venv/bin/python -r deploy/docker/requirements.txt`
- `git diff --check`

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas — N/A; the change is straightforward and the regression tests are named descriptively.
- [ ] I have made corresponding changes to the documentation — N/A; this restores documented Docker API behavior and explicitly leaves the remaining navigation bug tracked in the issue.

### DIFF
--- a/deploy/docker/requirements.txt
+++ b/deploy/docker/requirements.txt
@@ -14,3 +14,4 @@ PyJWT==2.10.1
 mcp>=1.18.0
 websockets>=15.0.1
 httpx[http2]>=0.27.2
+pypdf
--- a/deploy/docker/requirements.txt
+++ b/deploy/docker/requirements.txt
@@ -14,4 +14,4 @@ PyJWT==2.10.1
 mcp>=1.18.0
 websockets>=15.0.1
 httpx[http2]>=0.27.2
-pypdf
+pypdf>=6.0.0
--- a/tests/test_issue_2127_docker_pdf.py
+++ b/tests/test_issue_2127_docker_pdf.py
@@ -1,31 +1,71 @@
-import ast
+import importlib
 from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from packaging.requirements import InvalidRequirement, Requirement
+
+from crawl4ai.processors.pdf import PDFContentScrapingStrategy
 
 ROOT = Path(__file__).resolve().parent.parent
 
 
 def test_default_docker_dependencies_include_pypdf():
-    requirements = (
-        (ROOT / "deploy" / "docker" / "requirements.txt").read_text().splitlines()
-    )
+    lines = (ROOT / "deploy" / "docker" / "requirements.txt").read_text().splitlines()
+    names = set()
+    for line in lines:
+        line = line.strip()
+        if not line or line.startswith(("#", "-")):
+            continue
+        try:
+            names.add(Requirement(line).name)
+        except InvalidRequirement:
+            continue
 
-    assert "pypdf" in requirements
+    assert "pypdf" in names
 
 
-def test_stream_handler_preserves_requested_scraping_strategy():
-    tree = ast.parse((ROOT / "deploy" / "docker" / "api.py").read_text())
-    handler = next(
-        node
-        for node in tree.body
-        if isinstance(node, ast.AsyncFunctionDef)
-        and node.name == "handle_stream_crawl_request"
-    )
-    assigned_attributes = {
-        target.attr
-        for node in ast.walk(handler)
-        if isinstance(node, ast.Assign)
-        for target in node.targets
-        if isinstance(target, ast.Attribute)
+@pytest.mark.asyncio
+async def test_stream_handler_preserves_requested_scraping_strategy(monkeypatch):
+    docker_dir = ROOT / "deploy" / "docker"
+    monkeypatch.syspath_prepend(str(docker_dir))
+
+    api = importlib.import_module("api")
+    crawler_pool = importlib.import_module("crawler_pool")
+    egress_broker = importlib.import_module("egress_broker")
+    governor = importlib.import_module("governor")
+
+    crawler = MagicMock()
+    crawler.arun_many = AsyncMock(return_value=MagicMock())
+    monkeypatch.setattr(api, "_normalize_and_validate_seeds", lambda urls: urls)
+    monkeypatch.setattr(egress_broker, "enforce_egress", lambda _: None)
+    monkeypatch.setattr(governor, "clamp_deep_crawl", lambda _: None)
+    monkeypatch.setattr(crawler_pool, "get_crawler", AsyncMock(return_value=crawler))
+
+    crawler_config = {
+        "type": "CrawlerRunConfig",
+        "params": {
+            "cache_mode": "bypass",
+            "stream": False,
+            "scraping_strategy": {
+                "type": "PDFContentScrapingStrategy",
+                "params": {"extract_images": False, "batch_size": 8},
+            },
+        },
+    }
+    config = {
+        "crawler": {
+            "memory_threshold_percent": 90,
+            "rate_limiter": {"base_delay": [0.1, 0.3]},
+        }
     }
 
-    assert "scraping_strategy" not in assigned_attributes
+    await api.handle_stream_crawl_request(
+        urls=["https://example.com/document.pdf"],
+        browser_config={"type": "BrowserConfig", "params": {}},
+        crawler_config=crawler_config,
+        config=config,
+    )
+
+    effective_config = crawler.arun_many.await_args.kwargs["config"]
+    assert isinstance(effective_config.scraping_strategy, PDFContentScrapingStrategy)
--- a/tests/test_issue_2127_docker_pdf.py
+++ b/tests/test_issue_2127_docker_pdf.py
@@ -1,0 +1,31 @@
+import ast
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parent.parent
+
+
+def test_default_docker_dependencies_include_pypdf():
+    requirements = (
+        (ROOT / "deploy" / "docker" / "requirements.txt").read_text().splitlines()
+    )
+
+    assert "pypdf" in requirements
+
+
+def test_stream_handler_preserves_requested_scraping_strategy():
+    tree = ast.parse((ROOT / "deploy" / "docker" / "api.py").read_text())
+    handler = next(
+        node
+        for node in tree.body
+        if isinstance(node, ast.AsyncFunctionDef)
+        and node.name == "handle_stream_crawl_request"
+    )
+    assigned_attributes = {
+        target.attr
+        for node in ast.walk(handler)
+        if isinstance(node, ast.Assign)
+        for target in node.targets
+        if isinstance(target, ast.Attribute)
+    }
+
+    assert "scraping_strategy" not in assigned_attributes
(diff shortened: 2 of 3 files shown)


deploy/docker/requirements.txt
```diff
@@ -14,3 +14,4 @@ PyJWT==2.10.1
 mcp>=1.18.0
 websockets>=15.0.1
 httpx[http2]>=0.27.2
+pypdf
```
SohamKukreti:
```suggestion
pypdf>=6.0.0
```

tests/test_issue_2127_docker_pdf.py
```diff
@@ -0,0 +1,31 @@
...
+
+ROOT = Path(__file__).resolve().parent.parent
+
+
+def test_default_docker_dependencies_include_pypdf():
+    requirements = (
+        (ROOT / "deploy" / "docker" / "requirements.txt").read_text().splitlines()
+    )
+
+    assert "pypdf" in requirements
```
SohamKukreti: This exact-line match breaks as soon as the requirement gets a version spec (e.g. `pypdf>=6.0.0`, requested above) — the test would fail on a correct change. Parsing the requirement names makes it robust:

```python
from packaging.requirements import Requirement, InvalidRequirement

def test_default_docker_dependencies_include_pypdf():
    lines = (ROOT / "deploy" / "docker" / "requirements.txt").read_text().splitlines()
    names = set()
    for line in lines:
        line = line.strip()
        if not line or line.startswith(("#", "-")):
            continue
        try:
            names.add(Requirement(line).name)
        except InvalidRequirement:
            continue  # pip-specific syntax (inline comments, paths, etc.)
    assert "pypdf" in names
```

tests/test_issue_2127_docker_pdf.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    assigned_attributes = {
+        target.attr
+        for node in ast.walk(handler)
+        if isinstance(node, ast.Assign)
+        for target in node.targets
+        if isinstance(target, ast.Attribute)
+    }
+
+    assert "scraping_strategy" not in assigned_attributes
```
SohamKukreti: This asserts on the *shape of the source* rather than the behavior, which makes it both bypassable and over-strict — I verified both directions locally:

- It **passes with the bug reintroduced** via `setattr(crawler_config, "scraping_strategy", LXMLWebScrapingStrategy())` or via an annotated assignment (`crawler_config.scraping_strategy: object = ...` is an `ast.AnnAssign`, which this walk doesn't catch).
- It **fails on legitimate code**: a future default-only-if-unset pattern like `crawler_config.scraping_strategy = crawler_config.scraping_strategy or LXMLWebScrapingStrategy()` preserves client strategies but trips this assertion — so the test would actively obstruct the correct implementation if we ever want an explicit server-side default here.

A behavioral test would guard the actual contract. For example: load the issue's payload through `CrawlerRunConfig.load(..., provenance=Provenance.UNTRUSTED)`, run it through the same config-processing the stream handler does, and assert the config still holds a `PDFContentScrapingStrategy`. That stays green for any implementation that preserves the client's strategy and red for any that clobbers it, regardless of syntax.

SohamKukreti (dismissed):
Thanks @nightcityblade for the quick turnaround on this — both code changes are correct, and I've verified them against an image built from this branch. There's one scope issue that blocks merging as-is, detailed below; the smaller items are in inline comments.

  What I verified (image built from this branch, INSTALL_TYPE=default)

- ✅ pypdf is present in the default image; the ImportError from #2127 is gone on both /crawl and /crawl/stream.
- ✅ /crawl/stream now honors the client's scraping_strategy. Differential test (PDF strategy pointed at an HTML page): both endpoints now behave identically, where before the stream path silently swapped in LXMLWebScrapingStrategy and returned wrong content with no error.
- ✅ Removing the override causes no regression for clients that don't send a strategy: the library default has been LXMLWebScrapingStrategy since June 2025 (async_configs.py), so they get identical behavior — verified with plain crawls on both endpoints. The deleted line dates to Feb 2025 (392c923), when the library default was still the slower BS4 scraper; it's been redundant since the default changed, and its only remaining effect was this bug.
- ✅ Both new tests fail with the fix reverted; the branch merges cleanly into current develop.

Blocking: this PR should not close #2127

The issue's repro still fails on the image built from this branch — the error just changes from the ImportError to:

Failed on navigating ACS-GOTO: Page.goto: Download is starting

The container's headless Chromium can't render PDFs inline, so page.goto() turns every PDF URL into a download before the scraping strategy ever runs, and async_crawler_strategy.py only tolerates net::ERR_ABORTED. I tested 5 PDF hosts (arxiv, w3.org, pdfobject.com, irs.gov, raw.githubusercontent.com): 0/5 succeed. The accept_downloads escape hatch doesn't work through the API either (downloads_path is stripped by the untrusted-field allowlist, and the default /app/downloads isn't writable by appuser), and file:// URLs are blocked by SSRF protection by design — so there is currently no path to a working PDF scrape through the Docker server, with or without this PR.

Both changes here are necessary — they're just not sufficient for the issue's title claim. Two ways forward, either is fine with me:

1. Rescope: retitle to what the PR verifiably does — e.g. fix(docker): add pypdf and honor client scraping strategy on /crawl/stream — and reference #2127 without a closing keyword, leaving the navigation failure tracked in the issue (also related: #1367).
2. Extend: additionally make PDF navigation survivable — e.g. tolerate the download-triggered navigation failure when the configured scraping strategy is PDFContentScrapingStrategy, since that strategy fetches the PDF itself via its own download path and doesn't need the rendered page at all.

Also requested (see inline comments)

- Bound pypdf to match the library's existing spec in the root requirements.txt (pypdf>=6.0.0).
- Test robustness: parse requirement names instead of exact-line matching, and consider a behavioral test for the stream handler instead of the AST shape check — details and suggested code inline.

tests/test_issue_2127_docker_pdf.py
nightcityblade: Replaced the AST assertion in 10130de with an async behavioral test. It invokes the real `handle_stream_crawl_request`, deserializes the issue's `PDFContentScrapingStrategy` config through the handler, and verifies that the resulting strategy is passed to `crawler.arun_many`.

deploy/docker/requirements.txt
nightcityblade: Updated to `pypdf>=6.0.0` in 10130de, matching the library's root requirement.

tests/test_issue_2127_docker_pdf.py
nightcityblade: Updated in 10130de. The regression now parses each valid requirement with `packaging.requirements.Requirement`, so version constraints do not make a correct dependency entry fail the test.

nightcityblade:
Thanks for the thorough validation and for identifying the remaining Chromium navigation failure. I took the rescope option: the title and description no longer claim to close #2127, and the unresolved download-triggered navigation behavior is explicitly left tracked there (with #1367 noted as related).

The inline requests are addressed in 10130de: `pypdf>=6.0.0`, robust requirement parsing, and a real stream-handler behavioral regression. The focused stream/PDF suite passes (7 tests), with Black, `py_compile`, dependency resolution dry-run, and `git diff --check` all clean.

SohamKukreti:
I have opened PR #2150 which builds on top of this PR to support PDF scraping by using `AsyncWebCrawler(crawler_strategy=PDFCrawlerStrategy())` when `scraping_strategy` is `PDFContentScrapingStrategy`

PR #2131, opened 2026-08-08
fix(config): make body visibility timeout configurable

### LINKED ISSUE #2129: Hidden <body> adds a fixed 30s to every crawl, even when ignore_body_visibility=True (the default)
### Summary

`_crawl_web` waits for `document.body` to become visible using a hardcoded 30s timeout ([`async_crawler_strategy.py#L815-L827` on v0.9.2](https://github.com/unclecode/crawl4ai/blob/v0.9.2/crawl4ai/async_crawler_strategy.py#L815-L827)). The result is only acted on when `ignore_body_visibility=False` — and that flag **defaults to `True`**.

So on any page where the body never becomes visible, every crawl spends 30 seconds producing a value that is then discarded.

This isn't an exotic case: it is exactly what AngularJS `ng-cloak` (and Vue's `v-cloak`) leave behind whenever the app fails to bootstrap. The page still renders, the crawl still succeeds — it's just 30s slower, every single time.

### Reproduction

Self-contained, no external site needed. The two variants differ only by an `ng-cloak` attribute on `<body>`:

```python
import asyncio, threading, time
from crawl4ai import AsyncWebCrawler, BrowserConfig, CacheMode, CrawlerRunConfig

PROSE = "<p>Real, indexable prose content that the crawler should return.</p>" * 40
HIDDEN = f"""<!DOCTYPE html><html><head>
<style>[ng-cloak] {{ display: none; }}</style></head>
<body ng-cloak><h1>Hello</h1>{PROSE}</body></html>""".encode()
VISIBLE = HIDDEN.replace(b" ng-cloak", b"", 1)

class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        body = VISIBLE if self.path == "/visible" else HIDDEN
        self.send_response(200)
        self.send_header("Content-Type", "text/html")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)
    def log_message(self, *a): pass

async def main():
    threading.Thread(target=HTTPServer(("127.0.0.1", 8897), Handler).serve_forever,
                     daemon=True).start()
    await asyncio.sleep(0.3)
    async with AsyncWebCrawler(config=BrowserConfig(headless=True)) as crawler:
        for label, path in (("visible body", "/visible"), ("hidden body", "/hidden")):
            t = time.time()
            r = await crawler.arun(url=f"http://127.0.0.1:8897{path}",
                                   config=CrawlerRunConfig(cache_mode=CacheMode.BYPASS))
            print(f"  {label:<14} {time.time()-t:6.1f}s  success={r.success}  html={len(r.html or '')}b")

asyncio.run(main())
```

Output:

```
  visible body      0.3s  success=True  html=2834b
  hidden body      30.4s  success=True  html=2846b
```

### Real-world impact

On an Oracle B2C helpdesk we crawl, **every** URL took 34–35s. For comparison, from the same container: `curl` returned in 3.9s, and plain Playwright with the same `wait_until` in 4.7s. Instrumenting the pipeline attributed 30.1s of the 34.7s to `csp_compliant_wait`. Reference sites on that same container: 0.2s and 1.9s.

Reproduced independently on macOS and on a Linux server, so it is not environment-specific.

With the timeout lowered to 2000ms:

| page | default (30000) | 2000ms | extracted words |
|---|---|---|---|
| helpdesk home | 35.2s | 6.1s | 259 → 259 |
| helpdesk article | 34.2s | 6.5s | 962 → 962 |
| `intermedia.com` (visible body) | 1.8s | 1.0s | unaffected |
| `example.com` | 0.3s | 0.2s | unaffected |

Same content, ~5x faster.

### Suggested fix

Make the timeout configurable rather than changing behaviour — e.g. `CrawlerRunConfig.body_visibility_timeout`, defaulting to `30000` so existing behaviour is byte-for-byte unchanged. Callers who know their pages cloak the body can lower it.

I deliberately did **not** propose skipping the wait when `ignore_body_visibility=True`. I measured that variant too, and it returns *less* content (160 vs 259 words on the real page), because the wait doubles as render time. A configurable ceiling keeps that trade-off with the caller.

I have a patch (6 lines, all additive: the new field in `async_configs.py` plus the one call site) and verified against v0.9.2:

- `tests/test_config_defaults.py`: 33 passed, identical before and after
- serialisation round-trips (`to_dict`, `from_kwargs`, `dump`/`load`, `clone`) preserve the field
- default value reproduces current timing exactly (30.2s on the repro above)

Happy to open a PR if you'd like it.

### Environment

- crawl4ai 0.9.2 (verified in source) and 0.8.9 (Docker, where we hit it)
- Chromium via Playwright, headless
- macOS 15 and Debian container

nightcityblade:
## Summary

Fixes #2129.

Add `CrawlerRunConfig.body_visibility_timeout`, defaulting to the existing 30-second behavior, and use it for the body visibility check. This lets callers lower the ceiling for pages whose body remains hidden without changing the default crawl behavior. The value is included in config serialization and the untrusted-config allowlist, validated as a positive number, and capped at 60 seconds for untrusted Docker input.

- `crawl4ai/async_configs.py` - define, document, serialize, validate, allow, and cap the new configuration field.
- `crawl4ai/async_crawler_strategy.py` - pass the configured timeout to the body visibility wait.
- `docs/md_v2/api/parameters.md` and `docs/md_v2/complete-sdk-reference.md` - document the new public option.
- `tests/test_config_defaults.py` - cover the default and round trip, invalid values, the untrusted cap, and propagation into `csp_compliant_wait`.

## How Has This Been Tested?

- `/tmp/crawl4ai-followup-venv/bin/python -m pytest tests/test_config_defaults.py -q` — 40 passed, 1 existing warning.
- `/tmp/crawl4ai-followup-venv/bin/python -m pytest deploy/docker/tests/test_security_trust_boundary.py -q` — 38 passed, 4 existing warnings.
- `/tmp/crawl4ai-followup-venv/bin/ruff check --isolated --select E9,F63,F7,F82 --ignore F821 crawl4ai/async_configs.py crawl4ai/async_crawler_strategy.py tests/test_config_defaults.py` — passed. (`F821` is excluded because of the existing `VirtualScrollConfig` forward-reference finding in `async_crawler_strategy.py`.)
- `/tmp/crawl4ai-followup-venv/bin/python -m compileall -q crawl4ai/async_configs.py crawl4ai/async_crawler_strategy.py tests/test_config_defaults.py` — passed.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas — N/A; no complex logic was introduced.

### DIFF
--- a/crawl4ai/async_configs.py
+++ b/crawl4ai/async_configs.py
@@ -244,6 +244,7 @@ UNTRUSTED_FIELD_ALLOWLIST = {
         "fetch_ssl_certificate",
         # timing / waiting
         "wait_until", "page_timeout", "wait_for", "wait_for_timeout",
+        "body_visibility_timeout",
         "wait_for_images", "delay_before_return_html", "mean_delay", "max_range",
         # scrolling / rendering
         "ignore_body_visibility", "scan_full_page", "scroll_delay",
@@ -1463,6 +1464,8 @@ class CrawlerRunConfig():
                         Default: False.
         ignore_body_visibility (bool): If True, ignore whether the body is visible before proceeding.
                                        Default: True.
+        body_visibility_timeout (int): Maximum time in ms to wait for the body to become visible.
+                                       Default: 30000.
         scan_full_page (bool): If True, scroll through the entire page to load all content.
                                Default: False.
         scroll_delay (float): Delay in seconds between scroll steps if scan_full_page is True.
@@ -1640,6 +1643,7 @@ class CrawlerRunConfig():
         c4a_script: Union[str, List[str]] = None,
         js_only: bool = False,
         ignore_body_visibility: bool = True,
+        body_visibility_timeout: int = 30000,
         scan_full_page: bool = False,
         scroll_delay: float = 0.2,
         max_scroll_steps: Optional[int] = None,
@@ -1770,6 +1774,7 @@ class CrawlerRunConfig():
         self.c4a_script = c4a_script
         self.js_only = js_only
         self.ignore_body_visibility = ignore_body_visibility
+        self.body_visibility_timeout = body_visibility_timeout
         self.scan_full_page = scan_full_page
         self.scroll_delay = scroll_delay
         self.max_scroll_steps = max_scroll_steps
@@ -2137,6 +2142,7 @@ class CrawlerRunConfig():
             "js_code_before_wait": self.js_code_before_wait,
             "js_only": self.js_only,
             "ignore_body_visibility": self.ignore_body_visibility,
+            "body_visibility_timeout": self.body_visibility_timeout,
             "scan_full_page": self.scan_full_page,
             "scroll_delay": self.scroll_delay,
             "max_scroll_steps": self.max_scroll_steps,
--- a/crawl4ai/async_configs.py
+++ b/crawl4ai/async_configs.py
@@ -300,7 +300,7 @@ def _clamp_untrusted(type_name: str, params: dict) -> dict:
         return min(int(v), _MAX_TIMEOUT_MS)
 
     if type_name == "CrawlerRunConfig":
-        for f in ("page_timeout", "wait_for_timeout"):
+        for f in ("page_timeout", "wait_for_timeout", "body_visibility_timeout"):
             if f in params:
                 params[f] = _cap_timeout(params[f])
         if isinstance(params.get("max_scroll_steps"), int):
@@ -1774,6 +1774,12 @@ class CrawlerRunConfig():
         self.c4a_script = c4a_script
         self.js_only = js_only
         self.ignore_body_visibility = ignore_body_visibility
+        if (
+            not isinstance(body_visibility_timeout, (int, float))
+            or isinstance(body_visibility_timeout, bool)
+            or body_visibility_timeout <= 0
+        ):
+            raise ValueError("body_visibility_timeout must be a positive number")
         self.body_visibility_timeout = body_visibility_timeout
         self.scan_full_page = scan_full_page
         self.scroll_delay = scroll_delay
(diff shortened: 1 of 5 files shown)


ntohidi:
@nightcityblade Thanks — approach looks right. Default stays 30000 so nobody's existing crawls change, and I checked the plumbing: `from_kwargs` reads the `__init__` signature so that path is covered automatically, and `clone`/`dump`/`load` all go through `to_dict`, which you updated. Allowlist entry is in the right set too.

Three things before merge:

1. Docs. `docs/md_v2/api/parameters.md` and `docs/md_v2/complete-sdk-reference.md` both document `ignore_body_visibility` and should get the new option too. The docstring alone isn't enough.

2. The timeout goes straight into JavaScript at `async_crawler_strategy.py:324`. Passing `None` renders `> None` and blows up as a RuntimeError. That's pre-existing, but this PR makes it reachable from user input, including over the Docker API now that it's allowlisted. Worth a basic check on the value.

3. The test only checks the default and a dump/load round trip. Nothing verifies the value actually reaches `csp_compliant_wait` — the test would still pass if the call site kept using 30000. A mock asserting the timeout that gets passed would cover the real thing.

Separate from this PR: even after it lands, the default path still spends 30 silent seconds on these pages. `csp_compliant_wait` returns False on timeout instead of raising, so nothing is logged and the crawl reports success. @mvletter had to profile the pipeline to find it. A debug log when that wait times out would make this option discoverable — happy to do it as a follow-up rather than expand the scope here.

nightcityblade:
Thanks for the detailed review. Addressed in `64bf678`:

- documented `body_visibility_timeout` in both public parameter references;
- reject non-numeric, boolean, zero, and negative values, and cap untrusted Docker input at 60 seconds;
- added a mocked crawl-path test that verifies `1234` reaches `csp_compliant_wait`, alongside default/round-trip/validation coverage.

Focused validation passes: 40 config tests and 38 trust-boundary tests. I kept the separate timeout-debug logging idea out of this PR as suggested.

ntohidi:
@nightcityblade thanks for your contribution :)

PR #2137, opened 2026-08-12
fix: skip anti-bot detection for PDF responses

### LINKED ISSUE #2135: [Bug]: PDFCrawlerStrategy fails with "Blocked by anti-bot protection" -  the documented PDF example is broken
## crawl4ai version

0.9.2 (reproduced on latest `develop`, `ea26abb`)

## Expected Behavior

Running the exact example from the docs ([Advanced → PDF Parsing](https://docs.crawl4ai.com/advanced/pdf-parsing/), `docs/md_v2/advanced/pdf-parsing.md`) should return `result.success = True` with the extracted PDF content, as the page describes.

## Current Behavior

The documented `PDFCrawlerStrategy` + `PDFContentScrapingStrategy` pairing **always** returns `success: False` with:

```
Blocked by anti-bot protection: Near-empty content (33 bytes) with HTTP 200
```

— on every PDF, every time. The extraction itself actually succeeds: the same `CrawlResult` carries the complete PDF markdown (25,297 chars for the docs' example URL). The result is self-contradictory — full content plus a failure flag — so any caller that checks `result.success` (as the docs example does) discards a perfectly good scrape.

## Root cause

1. `PDFCrawlerStrategy.crawl()` returns placeholder HTML — literally `"Scraper will handle the real work"`, 33 bytes (`crawl4ai/processors/pdf/__init__.py:17`). The real extraction happens later in the scraping strategy, which populates `cleaned_html`/markdown from the PDF.
2. `is_blocked()` only ever sees this pre-scrape placeholder, never the scraped output. Verified by instrumenting `is_blocked()`: it fires twice per crawl, both times with the 33-byte stub —
   - once inside the attempt loop (`crawl4ai/async_webcrawler.py:512`), which classifies the attempt as *blocked* — so with retries/`proxy_rotation_strategy` configured, every proxy and retry is burned on a crawl that is actually succeeding, and a configured `fallback_fetch_function` gets invoked pointlessly;
   - once in the final post-processing veto (`crawl4ai/async_webcrawler.py:629`), which is what flips `success = False`.
3. 33 stripped bytes with HTTP 200 trips the near-empty-content heuristic (`crawl4ai/antibot_detector.py:274`).

Notably, the code already exempts three cases where this heuristic misfires — successful fallback fetches, `raw:` URLs, and binary downloads ("`is_blocked()` would misread '0 bytes html' as a block", `async_webcrawler.py:619-628`). A `PDFCrawlerStrategy` crawl is a fourth case of the same shape: `html` is a stub by design, and the content lives elsewhere. It matches none of the existing exemptions (`downloaded_files` is not set; the PDF goes through a temp file).

## Is this reproducible?

Yes — deterministic, 100% of PDFs. Also reproduces with local `file://` PDFs.

## Inputs Causing the Bug

Any `arun()` call through `AsyncWebCrawler(crawler_strategy=PDFCrawlerStrategy())`.

## Steps to Reproduce

1. `pip install crawl4ai[pdf]`
2. Run the example from the PDF parsing docs page, unmodified.
3. Observe `Failed to process PDF: Blocked by anti-bot protection: Near-empty content (33 bytes) with HTTP 200` — while `result.markdown.raw_markdown` contains the full extracted paper.

## Code snippets

```python
import asyncio
from crawl4ai import AsyncWebCrawler, CrawlerRunConfig
from crawl4ai.processors.pdf import PDFCrawlerStrategy, PDFContentScrapingStrategy

async def main():
    # Exact example from docs/md_v2/advanced/pdf-parsing.md
    pdf_crawler_strategy = PDFCrawlerStrategy()
    pdf_scraping_strategy = PDFContentScrapingStrategy()
    run_config = CrawlerRunConfig(scraping_strategy=pdf_scraping_strategy)

    async with AsyncWebCrawler(crawler_strategy=pdf_crawler_strategy) as crawler:
        pdf_url = "https://arxiv.org/pdf/2310.06825.pdf"
        result = await crawler.arun(url=pdf_url, config=run_config)
        if result.success:
            print(f"Successfully processed PDF: {result.url}")
        else:
            print(f"Failed to process PDF: {result.error_message}")
            md = result.markdown.raw_markdown if hasattr(result.markdown, "raw_markdown") else result.markdown
            print(f"...yet extracted markdown length is: {len(md or '')}")

```

Output:

```
[SCRAPE].. ◆ https://arxiv.org/pdf/2310.06825.pdf  | ✓ | ⏱: 1.07s
[ERROR]... × https://arxiv.org/pdf/2310.06825.pdf  | Error: Blocked by anti-bot protection: Near-empty content (33 bytes) with HTTP 200
Failed to process PDF: Blocked by anti-bot protection: Near-empty content (33 bytes) with HTTP 200
...yet extracted markdown length is: 25297
```

## OS

Linux

## Python version

3.10

## Related

- #2058 — general request for a config option to disable the `is_blocked()` post-crawl veto (this report is a deterministic instance of that mechanism misfiring on a documented first-party workflow)
- Workaround for anyone hitting this: use `AsyncWebCrawler(crawler_strategy=AsyncHTTPCrawlerStrategy())` with `PDFContentScrapingStrategy` — verified working (full extraction, `success: True`), since the HTTP strategy returns the real PDF bytes as the response body and the heuristic never fires.

nightcityblade:
## Summary

Fixes #2135.

Treat responses declared as `application/pdf` like other binary content before anti-bot classification. This prevents `PDFCrawlerStrategy`'s intentional 33-byte HTML placeholder from triggering retries, fallback fetching, and the final failure veto after PDF extraction succeeds.

- `crawl4ai/async_webcrawler.py` - detect PDF response content types and bypass HTML-specific anti-bot checks in both classification stages.
- `tests/test_issue_2135_pdf_antibot.py` - exercise the real `PDFCrawlerStrategy` through `AsyncWebCrawler.arun()` and verify one attempt, success, and no fallback.

## How Has This Been Tested?

- `.venv/bin/python -m pytest tests/test_issue_2135_pdf_antibot.py -q` - 1 passed.
- `.venv/bin/python tests/proxy/test_antibot_detector.py` - 47 passed, 0 failed.
- `.venv/bin/ruff check --isolated --select E9,F63,F7,F82 --ignore F821 crawl4ai/async_webcrawler.py tests/test_issue_2135_pdf_antibot.py` - passed.
- `.venv/bin/black --check --target-version py312 --fast tests/test_issue_2135_pdf_antibot.py` - passed.
- `.venv/bin/python -m compileall -q crawl4ai/async_webcrawler.py tests/test_issue_2135_pdf_antibot.py` - passed.
- `git diff --check upstream/develop...HEAD` - passed.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas - N/A; the PDF-response condition is explicit and covered by a focused regression test.
- [ ] I have made corresponding changes to the documentation - N/A; this fixes the implementation to match the existing PDF parsing documentation.
- [x] New and existing relevant tests pass locally with my changes

### DIFF
--- a/crawl4ai/async_webcrawler.py
+++ b/crawl4ai/async_webcrawler.py
@@ -408,6 +408,7 @@ class AsyncWebCrawler:
                     _block_reason = ""
                     _done = False
                     crawl_result = None
+                    _is_pdf_response = False
                     _crawl_stats = {
                         "attempts": 0,
                         "retries": 0,
@@ -458,6 +459,13 @@ class AsyncWebCrawler:
 
                                 async_response = await self.crawler_strategy.crawl(
                                     url, config=config)
+                                _is_pdf_response = any(
+                                    key.lower() == "content-type"
+                                    and "application/pdf" in str(value).lower()
+                                    for key, value in (
+                                        async_response.response_headers or {}
+                                    ).items()
+                                )
 
                                 html = sanitize_input_encode(async_response.html)
                                 screenshot_data = async_response.screenshot
@@ -505,7 +513,7 @@ class AsyncWebCrawler:
 
                                 # Check if blocked (skip for raw: URLs —
                                 # caller-provided content, anti-bot N/A)
-                                if _is_raw_url:
+                                if _is_raw_url or _is_pdf_response:
                                     _blocked = False
                                     _block_reason = ""
                                 else:
@@ -625,7 +633,12 @@ class AsyncWebCrawler:
                         # empty by design, and is_blocked() would misread "0 bytes
                         # html" as a block.
                         _has_download = bool(getattr(crawl_result, "downloaded_files", None))
-                        if not _fallback_succeeded and not _is_raw_url and not _has_download:
+                        if (
+                            not _fallback_succeeded
+                            and not _is_raw_url
+                            and not _has_download
+                            and not _is_pdf_response
+                        ):
                             _blocked, _block_reason = is_blocked(
                                 crawl_result.status_code, crawl_result.html or "")
                             if _blocked:
--- a/tests/test_issue_2135_pdf_antibot.py
+++ b/tests/test_issue_2135_pdf_antibot.py
@@ -1,0 +1,30 @@
+import pytest
+
+from crawl4ai import AsyncWebCrawler, CacheMode, CrawlerRunConfig, CrawlResult
+from crawl4ai.processors.pdf import PDFCrawlerStrategy
+
+
+@pytest.mark.asyncio
+async def test_pdf_response_skips_antibot_retries_and_fallback():
+    fallback_calls = []
+
+    async def process_html(url, html, **kwargs):
+        return CrawlResult(url=url, html=html, success=True, status_code=200)
+
+    async def fallback(url):
+        fallback_calls.append(url)
+        return "<html>fallback</html>"
+
+    crawler = AsyncWebCrawler(crawler_strategy=PDFCrawlerStrategy())
+    crawler.aprocess_html = process_html
+    result = await crawler.arun(
+        "https://example.com/document.pdf",
+        config=CrawlerRunConfig(
+            cache_mode=CacheMode.BYPASS,
+            max_retries=1,
+            fallback_fetch_function=fallback,
+        ),
+    )
+    assert result.success
+    assert result.crawl_stats["attempts"] == 1
+    assert fallback_calls == []


nightcityblade:
CI note: the only failing check is the repository's `notify-discord` job because `DISCORD_WEBHOOK` is empty for this fork PR. The job exits before any code validation; the focused PDF/anti-bot tests and lint checks reported in the PR body are green.

ntohidi:
Thanks @nightcityblade — your fix was correct and caught both call sites. We went with #2138 instead because the `placeholder_html` flag lets any strategy self-declare a stub response, so we don't need to edit these guards again for the next non-HTML strategy. Nothing wrong with your approach; it was a design preference. Appreciate the quick turnaround on this one.

PR #2142, opened 2026-08-15
Docker: Chain egress proxy through upstream HTTP(S)_PROXY

SohamKukreti:
## Summary
Since 0.9.0, the Docker server's egress pinning proxy dials target IPs directly (`asyncio.open_connection`), ignoring `HTTP_PROXY`/`HTTPS_PROXY`. On hosts that can only reach the internet through a corporate proxy, every crawl fails with `Page.goto: net::ERR_TIMED_OUT` — this worked in 0.8.x. Reported in discussion #2041.

This PR teaches the pinning proxy to chain through the upstream proxy **without weakening the SSRF/rebinding guarantees**: validation (`resolve_and_pin`) still runs locally *before* the upstream is ever contacted, and the upstream is asked to `CONNECT <pinned-ip>:<port>` — never a hostname — so it performs no DNS resolution of its own and the pin holds.

**Behavior:**
- Reads standard `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` (both cases; scheme-aware selection with fallback); `CRAWL4AI_UPSTREAM_PROXY` overrides them.
- `NO_PROXY` supports domain-suffix (`.corp.lan`) and IP/CIDR (`192.168.0.0/16`) entries — internal targets should be listed there to dial direct.
- Basic auth via `http://user:pass@proxy:port` (percent-decoded, UTF-8) sent preemptively; on the CONNECT path a `407`/non-`200` from the upstream fails fast with the existing opaque `403` plus a server-side log line, while on plain HTTP the upstream's error response is relayed to the browser as-is.
- Plain-HTTP requests are re-issued to the upstream in absolute form against the pinned IP with `Connection: close` forced, so a reused client connection cannot smuggle unvalidated requests past the pin.
- With no proxy env vars set, the dial path is byte-identical to current behavior — zero change for existing deployments.

**Known limitations (documented in the README):**
- Proxies that refuse CONNECT-to-an-IP (hostname/domain ACLs) and containers with no DNS at all are not supported yet (a future opt-in hostname mode could cover them).
- NTLM/Kerberos/SPNEGO proxies are not supported — front them with a local translator (e.g. `cntlm`, `px`) and point `CRAWL4AI_UPSTREAM_PROXY` at it.

- `deploy/docker/egress_proxy.py` — upstream detection (`upstream_proxy()`, `_no_proxy_match()`, `_use_upstream()`), the chained `_dial()` (CONNECT-to-pinned-IP, auth, header drain so upstream headers never leak into the tunnel), absolute-form + `Connection: close` for chained plain HTTP, startup/refusal log lines.
- `deploy/docker/tests/test_security_egress_proxy.py` — 3 new tests covering the security contract: chained CONNECT carries only the pinned IP and blocked targets produce zero upstream traffic; plain-HTTP absolute-form + anti-smuggling; env parsing (precedence, scheme selection, credentials, `NO_PROXY` routing). Plus an autouse fixture clearing proxy env vars so the suite is deterministic on dev machines behind proxies.
- `deploy/docker/README.md` — "Behind a corporate proxy" note in the Run the Container section (env vars, auth support, limitations).

## How Has This Been Tested?
- Full Docker security suite: **319 passed, 1 xfailed** (includes the 0.9 hardening contract tests, all unchanged).
- Library regression suite (`tests/regression/`, non-network): **293 passed**; the single failure (`test_cosine_basic`, missing optional `transformers`) was verified pre-existing by re-running on the unmodified tree.
- Adversarial pass (32 targeted tests, real sockets, no mocking): env parsing edge cases (malformed/whitespace URLs, non-latin-1 passwords), `NO_PROXY` matching (anchored suffixes, CIDR, the exact list from #2041), dead/407/garbage upstreams (fast opaque 403, no hang), IPv6 pin bracketing, 8 concurrent tunnels with no cross-talk, and the smuggling regression.
- End-to-end: real Chromium → pinning proxy → a real forwarding "corporate" proxy → origin server, reproducing the #2041 topology — page + subresources each arrive as separate validated pinned-IP requests.
- Proxy-auth probe: preemptive Basic accepted first-try by an auth-requiring proxy; wrong/missing creds and NTLM-only proxies fail fast (single attempt, opaque 403, warning logged).

## Checklist:

### DIFF
--- a/deploy/docker/README.md
+++ b/deploy/docker/README.md
@@ -116,6 +116,17 @@ EOL
 
 > The server will be available at `http://localhost:11235`. Visit `/playground` to access the interactive testing interface.
 
+*   **Behind a corporate proxy:** if the host reaches the internet only through
+    an HTTP proxy, set the standard `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY`
+    env vars (Docker's `proxies` config injects them automatically) — the
+    server's egress proxy chains through it while keeping its SSRF protections
+    (the upstream is asked to CONNECT to an already-validated, pinned IP).
+    `CRAWL4AI_UPSTREAM_PROXY` overrides the env vars. Basic auth via
+    `http://user:pass@proxy:port` is supported; for NTLM/Kerberos proxies,
+    front them with a local translator (e.g. `cntlm`, `px`) and point
+    `CRAWL4AI_UPSTREAM_PROXY` at it. Proxies that refuse CONNECT-to-an-IP, or
+    containers with no DNS at all, are not yet supported.
+
 #### 4. Stopping the Container
 
 ```bash
--- a/deploy/docker/egress_proxy.py
+++ b/deploy/docker/egress_proxy.py
@@ -13,13 +13,21 @@ TLS stays end-to-end (we tunnel ciphertext; Chromium verifies the cert/SNI
 against the real host - no MITM).
 
 Bound to 127.0.0.1 on an ephemeral port; started at server boot.
+
+If HTTP_PROXY/HTTPS_PROXY (or CRAWL4AI_UPSTREAM_PROXY) is set, we still
+resolve-and-pin locally but dial via the upstream proxy, asking it to CONNECT
+to the PINNED IP — never the hostname — so the rebinding guarantee holds.
+NO_PROXY bypasses it; with no proxy env set, behavior is unchanged.
 """
 
 from __future__ import annotations
 
 import asyncio
+import base64
+import ipaddress
 import logging
-from urllib.parse import urlsplit
+import os
+from urllib.parse import unquote, urlsplit
 
 from egress_broker import EgressBlocked, resolve_and_pin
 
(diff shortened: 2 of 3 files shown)


fchinch:
Thanks for this — the chaining works exactly as described for the standard
"container can only reach the internet through a corporate proxy" case, and the
`CONNECT <pinned-ip>` design keeps the rebinding guarantee intact. Nice.

I wanted to flag a related deployment shape that this doesn't cover yet, in case
it's useful input for the opt-in hostname mode you mention under known limitations.

**Our setup:** the proxy isn't an egress proxy — it's the entry point of a reverse
tunnel. An agent runs inside the customer's private network and dials out to our
hub; the hub exposes an HTTP CONNECT endpoint. When the browser CONNECTs, the hub
forwards the request over that tunnel and **the agent performs the DNS resolution
and the connection from inside the customer's network**. That's the whole point:
it lets us crawl sites that only exist there.

**Why the PR doesn't reach that case**, all three by design rather than oversight:

1. `resolve_and_pin` runs locally first, and an internal hostname doesn't resolve
   from our side at all.
2. Even when it does resolve, the answer is RFC1918 and is rejected before the
   upstream is considered.
3. The upstream receives `CONNECT <pinned-ip>:<port>` — but the far side needs the
   **hostname**, because resolution is supposed to happen over there. An IP we
   resolved locally is either wrong or meaningless in the target network.

`NO_PROXY` points the other way too: it treats internal targets as things to dial
direct, whereas for us internal targets are precisely what must traverse the proxy.

**What would unblock it** is the opt-in hostname mode you already floated: when an
operator explicitly configures an upstream and opts in, send `CONNECT <hostname>`
and let the upstream resolve, skipping the local pin for that path. That is a
deliberate trust decision — you're trusting a proxy you configured — so gating it
behind an explicit env var (rather than inferring it from `HTTP_PROXY`) seems right.

Happy to test a branch against our stack if you build one; we have a reproducible
environment with the tunnel and a private-network target, and can report back with
logs.

fchinch:
@SohamKukreti @unclecode @ntohidi  I added this PR:
https://github.com/unclecode/crawl4ai/pull/2152

ntohidi (changes requested):
Nice work — the security contract holds. I checked the part that matters: `resolve_and_pin` runs before the upstream is ever contacted, the upstream only ever gets `CONNECT <ip>:<port>` (never a name, so it has nothing to resolve), its reply headers are drained so they can't leak into the tunnel, and with no proxy env set the dial path is unchanged. Ran the suite locally — the 3 new tests pass.

Two small things first, both in `upstream_proxy()` (`deploy/docker/egress_proxy.py:45`). Neither breaks anyone who works today, but both fail *silently*, which is the same mystery timeout that got #2041 opened.

**1. A junk value hides a good one.** `_env` takes the first var that is *set*, then we bail if it won't parse — so a valid fallback never gets a look:

```
HTTPS_PROXY="   "      + HTTP_PROXY="http://good:3128"  ->  None
HTTPS_PROXY="http://"  + HTTP_PROXY="http://good:3128"  ->  None
```

Result: proxy silently off, every crawl times out. Fix: keep the first candidate that *parses*, not the first that exists.

**2. An `https://` proxy URL is mis-dialed.**

```
HTTPS_PROXY=https://proxy.corp  ->  ('proxy.corp', 80, None)
```

Port should be 443, and we open a plain socket rather than TLS. A `logger.warning` (or an outright refusal) would be enough — right now the operator gets a timeout and no clue.

Two nits, take them or leave them:

- `NO_PROXY=site.example:80` never matches — the port breaks the suffix test, and real `NO_PROXY` lists often carry ports.
- The description says a `407` "fails fast with the existing opaque 403". That's the CONNECT path only; on plain HTTP the upstream `407` is spliced straight back to Chromium. Not a security issue, just worth correcting in the text.

Happy to merge once 1 and 2 are in.

SohamKukreti:
All four addressed in 01e6e96: env parsing now keeps the first candidate that parses (junk and unsupported schemes fall through with a warning — including unparseable values, since they fail the same silent way), `NO_PROXY` host:port entries match, and the 407 sentence in the description is corrected to be CONNECT-path-only.
Thanks!

ntohidi (approved):
Verified 01e6e96 locally — all four addressed, no regressions.

Junk and unsupported-scheme values now fall through to a valid fallback with a warning, `https://`/`socks5://` are refused rather than silently dialed as plaintext on port 80, and `NO_PROXY` host:port entries match. Ran 24 checks over env parsing, precedence, auth (incl. non-latin-1), and 10 `NO_PROXY` shapes — all pass, and the near-miss suffix case still correctly does not match. Suite green.

The security path (`_dial`, `_use_upstream`, CONNECT handling) is untouched by the follow-up, so the original review stands. Thanks for the quick turnaround.

PR #2145, opened 2026-08-17
fix(crawler): warn when the body-visibility wait times out

### LINKED ISSUE #2144: Body-visibility wait times out silently, making body_visibility_timeout undiscoverable
### Summary

When the body-visibility wait times out, nothing is logged. The crawl reports `success=True` with an empty `error_message`, and 30 seconds vanish with no indication of where they went.

Follow-up to #2129 / #2131. That PR added `CrawlerRunConfig.body_visibility_timeout` so callers can lower the ceiling — but there is currently no way for a user to discover they need it.

### Why this matters

`csp_compliant_wait` returns `False` on timeout rather than raising (`crawl4ai/async_crawler_strategy.py:335-342`). On the default path the return value is discarded, because `ignore_body_visibility` defaults to `True`:

```python
is_visible = await self.csp_compliant_wait(page, """...""", timeout=config.body_visibility_timeout)

if not is_visible and not config.ignore_body_visibility:   # never taken by default
    ...
```

So the timeout is completely silent. @mvletter had to instrument the pipeline to attribute the delay — that is a high bar for what turns out to be a one-line config change.

### Measured on `develop`

Two identical pages differing only by `ng-cloak` on `<body>`:

```
  visible body      0.2s  success=True  words=362
  hidden body      30.3s  success=True  words=362
```

Instrumented:

```
total crawl:            30.4s
csp_compliant_wait:     30.1s  (timeout=30000, returned False)
ignore_body_visibility: True  -> wait result is DISCARDED
crawl reported success: True, error_message=''
```

Same content, 150x slower, no signal of any kind.

### Suggested fix

Log a warning (or debug, if warning is judged too noisy) when the body-visibility wait times out, naming the option:

> Body never became visible after 30000ms — the page may use `ng-cloak`/`v-cloak`. Lower `body_visibility_timeout` to speed up crawls of this site.

Scope is small: one branch at the `csp_compliant_wait` call site in `_crawl_web` (`crawl4ai/async_crawler_strategy.py:815-830`). No behavior change, no new config.

### Note

Lowering `body_visibility_timeout` does not help on pages served with a CSP `sandbox` directive. The timeout is enforced by a JS polling loop inside the page (`async_crawler_strategy.py:315-332`), and `sandbox` disables timers, so the loop's clock never advances — same class of problem as #2139. Requires both a sandboxed page and a hidden body, so it is narrow, but a log line would surface that case too.

ntohidi:
## Summary

Fixes #2144.

When the body-visibility wait times out, nothing is logged. The crawl reports `success=True` with an empty `error_message`, and `body_visibility_timeout` ms disappear with no indication of where they went — because the wait's result is discarded whenever `ignore_body_visibility` is `True` (the default).

That silence is what made the delay in #2129 impossible to attribute without instrumenting the pipeline, and what makes `body_visibility_timeout` (added in #2131) undiscoverable by exactly the users who need it.

This warns once when the wait times out **and** its result is ignored, naming the option and the timeout that actually applied. The strict path (`ignore_body_visibility=False`) already raises with visibility details, so it stays quiet — no double reporting.

No config, no behavior change, no new dependency. One log line.

- `crawl4ai/async_crawler_strategy.py` — warn after the visibility wait when it timed out and the result is about to be discarded.
- `tests/test_body_visibility_warning.py` — new: warns and reports the applied timeout; silent when the body is visible; silent on the strict path.

## How Has This Been Tested?

Live output on the #2129 repro (two identical pages differing only by `ng-cloak` on `<body>`) — warning on the hidden page, silence on the visible one:

```
[COMPLETE] ● http://127.0.0.1:8897/visible  | ✓ | ⏱: 0.34s
[WARNING]. ⚠ Body never became visible after 30000ms — the page may use ng-cloak/v-cloak.
             This delay is added to every crawl of this page; lower
             CrawlerRunConfig.body_visibility_timeout to shorten it.
[COMPLETE] ● http://127.0.0.1:8897/hidden   | ✓ | ⏱: 30.25s
```

- `tests/test_body_visibility_warning.py` + `tests/test_config_defaults.py` — 43 passed.
- Fail-on-revert verified: with the change reverted, `test_warns_when_wait_times_out_and_result_is_discarded` fails.
- `tests/regression/` — 320 passed, 2 failed. Both fail identically on `develop` without this change: `test_cosine_basic` is an ImportError from the missing optional `transformers` extra, and `test_soft_404_filters_probes` is pre-existing.

## Checklist:

- [ ] I have made corresponding changes to the documentation — N/A; no config or behavior change, only a log line.

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -826,6 +826,21 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
                     timeout=config.body_visibility_timeout,
                 )
 
+                if not is_visible and config.ignore_body_visibility:
+                    # The wait timed out and its result is about to be discarded,
+                    # so the crawl still succeeds — just body_visibility_timeout ms
+                    # slower. Say so, otherwise the delay is invisible (see #2144).
+                    self.logger.warning(
+                        message=(
+                            "Body never became visible after {timeout}ms — the page may "
+                            "use ng-cloak/v-cloak. This delay is added to every crawl of "
+                            "this page; lower CrawlerRunConfig.body_visibility_timeout "
+                            "to shorten it."
+                        ),
+                        tag="WARNING",
+                        params={"timeout": config.body_visibility_timeout},
+                    )
+
                 if not is_visible and not config.ignore_body_visibility:
                     visibility_info = await self.check_visibility(page)
                     raise Error(f"Body element is hidden: {visibility_info}")
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -812,6 +812,7 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
                 await page.wait_for_selector("body", state="attached", timeout=30000)
 
                 # Use the new check_visibility function with csp_compliant_wait
+                _visibility_wait_start = time.perf_counter()
                 is_visible = await self.csp_compliant_wait(
                     page,
                     """() => {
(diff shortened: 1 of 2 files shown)


ntohidi:
Addressed all three review findings in `1bac351`.

**2 — invisible where it matters most.** Confirmed and worse than described: `AsyncLogger._log` gates on `self.verbose or force_verbose` (`async_logger.py:238`), *and* `arun()` overwrites the logger's verbose from `CrawlerRunConfig.verbose` at `async_webcrawler.py:257`. So a run config with `verbose=False` suppressed the warning regardless of how the crawler was built. Verified end to end on the #2129 repro — before this commit the hidden-body page produced 30s of total silence with `verbose=False`; now it warns. Added `force_verbose=True` and a test asserting it.

**1 — warning can be false.** Fixed by timing the wait and only warning when it actually consumed its budget (>= 90% of `body_visibility_timeout`). An eval failure returns `False` immediately, so it no longer gets blamed on the timeout. The message now reports elapsed time rather than the configured value, so the number matches the delay the user is trying to explain (`after 30093ms`). Added a test for the early-failure path.

**3 — brittle assertion.** Now selects the warning call by message instead of reading the last call.

Tests: 45 passed (5 in this file, up from 3, plus `test_config_defaults.py`). `tests/regression/` — 320 passed, 2 failed, both failing identically on `develop` without this branch (`test_cosine_basic` is a missing optional `transformers` extra; `test_soft_404_filters_probes` is pre-existing).

PR #2160, opened 2026-08-20
Clean up leaked Playwright driver when browser launch fails in `__aenter__`

### LINKED ISSUE #2155: [Bug]: AsyncWebCrawler leaks Playwright driver process when browser startup fails in __aenter__
### crawl4ai version

0.8.0

### Expected Behavior

If browser startup fails while entering `AsyncWebCrawler`, all resources that were partially initialized during `__aenter__()` / `start()` should be cleaned up.

In particular, the Playwright `node ... cli.js run-driver` subprocess should exit before the exception is propagated.

### Current Behavior

When Playwright starts successfully but Chromium launch fails, `AsyncWebCrawler.__aenter__()` propagates the exception without closing the partially initialized crawler.

Each failed attempt leaves one live direct child process:

```text
.../playwright/driver/node .../playwright/driver/package/cli.js run-driver
```

Repeated failures therefore grow the process count linearly:

```text
attempt=1 playwright_drivers=1
attempt=2 playwright_drivers=2
attempt=3 playwright_drivers=3
```

On Linux/Python 3.11, each leaked child can also retain an `asyncio-waitpid-*` watcher thread until that child exits.

The relevant lifecycle currently looks like:

```python
async def __aenter__(self):
    return await self.start()
```

and browser startup initializes Playwright before launching Chromium:

```python
self.playwright = await async_playwright().start()
self.browser = await self.playwright.chromium.launch(**browser_args)
```

If the second line raises, `__aenter__()` never completes, so Python does not call `__aexit__()`. The already-started Playwright driver is not rolled back.

### Is this reproducible?

Yes

### Inputs Causing the Bug

- Any failure after Playwright has started but before browser startup completes.
- The reproducer below makes this deterministic by pointing `PLAYWRIGHT_BROWSERS_PATH` at an empty temporary directory.

### Steps to Reproduce

1. Install Crawl4AI 0.8.0.
2. Save and run the script below.
3. Observe that the number of Playwright driver child processes grows from 1 to 3.
4. The script terminates only the child processes it created before exiting.

### Code snippets

```python
import asyncio
import os
import tempfile
import time

import psutil

# Force BrowserType.launch() to fail after the Playwright driver has started.
empty_browsers = tempfile.mkdtemp(prefix="crawl4ai-empty-browsers-")
os.environ["PLAYWRIGHT_BROWSERS_PATH"] = empty_browsers

from crawl4ai import AsyncWebCrawler, BrowserConfig  # noqa: E402


parent = psutil.Process()
initial_pids = {p.pid for p in parent.children(recursive=True)}


def new_children():
    return [
        p
        for p in parent.children(recursive=True)
        if p.pid not in initial_pids
    ]


def playwright_drivers():
    result = []
    for process in new_children():
        try:
            command = " ".join(process.cmdline())
        except (psutil.NoSuchProcess, psutil.AccessDenied):
            continue
        if "playwright/driver" in command and "run-driver" in command:
            result.append(process)
    return result


async def fail_during_enter():
    async with AsyncWebCrawler(config=BrowserConfig(headless=True)):
        pass


try:
    for attempt in range(1, 4):
        try:
            asyncio.run(fail_during_enter())
        except Exception as exc:
            print(f"attempt={attempt} error={type(exc).__name__}: {exc}")

        time.sleep(0.5)
        drivers = playwright_drivers()
        print(
            f"attempt={attempt} playwright_drivers={len(drivers)} "
            f"pids={[p.pid for p in drivers]}"
        )
finally:
    # Do not leave processes behind after running the reproducer.
    children = new_children()
    for process in reversed(children):
        try:
            process.terminate()
        except psutil.NoSuchProcess:
            pass

    _, alive = psutil.wait_procs(children, timeout=3)
    for process in alive:
        try:
            process.kill()
        except psutil.NoSuchProcess:
            pass
    psutil.wait_procs(alive, timeout=3)
```

### Supporting Information

An application-level control using explicit lifecycle cleanup does not leak:

```python
crawler = AsyncWebCrawler(config=BrowserConfig(headless=True))
try:
    await crawler.start()
finally:
    await crawler.close()
```

With the same forced launch failure, three attempts produce:

```text
attempt=1 playwright_drivers=0
attempt=2 playwright_drivers=0
attempt=3 playwright_drivers=0
```

A possible library-level fix would be to make partial startup exception-safe, ideally in `BrowserManager.start()`, and add a regression test asserting that a failed browser launch leaves no Playwright driver child process:

```python
try:
    self.playwright = await async_playwright().start()
    self.browser = await self.playwright.chromium.launch(**browser_args)
except BaseException:
    await self.close()
    raise
```

Related but not identical: #1242 reports an exception through the same `AsyncWebCrawler.__aenter__() -> BrowserManager.start()` path followed by `BaseSubprocessTransport` cleanup against a closed event loop.

### OS

macOS 15.6.1, arm64

### Python version

3.11.1

### Browser

Chromium (launch intentionally forced to fail)

### Browser version

N/A

Representative error:

```text
playwright._impl._errors.Error: BrowserType.launch:
Executable doesn't exist at <empty PLAYWRIGHT_BROWSERS_PATH>

Loop <_UnixSelectorEventLoop running=False closed=True debug=False>
that handles pid <pid> is closed
```

SohamKukreti:
## Summary
Fixes #2155

When `AsyncWebCrawler.__aenter__()` fails partway through browser startup, the already-started Playwright driver process (`node .../cli.js run-driver`) was never cleaned up. Python does not call `__aexit__` for a failed `__aenter__`, so nothing stopped the driver — every failed startup attempt leaked one live child process. Retry loops made the process count grow linearly.

## Root cause

`BrowserManager.start()` initializes Playwright first and launches the browser later, with no error handling in between:

```python
self.browser = await self.playwright.chromium.launch(**browser_args)  # can raise
```

Any failure after the first line (missing browser executable, broken WebKit deps, unreachable CDP endpoint, task cancellation) propagated out of `__aenter__` and left the driver running.

## Fix

Wrap the startup body of `BrowserManager.start()` in `try/except BaseException`:

- on any failure, roll back partial state via `close()` before re-raising
- for external-CDP configs, `close()` intentionally returns early without stopping the driver (`cdp_cleanup_on_close=False` default), so the rollback additionally stops the driver explicitly — we own it even when we don't own the browser
- cleanup errors are suppressed so the **original** launch error always propagates, unchanged (bare `raise`)
- `BaseException` (not `Exception`) so cancellation and Ctrl-C during startup also clean up

> **Note for reviewers:** the diff looks large because the existing `start()` body moved one indent level into the `try` block. Review with `git diff -w` (or `-b`) — the real change is **15 added lines, zero modified lines**.

## Reproduction

Verified with the issue's reproducer and with a natural failure (WebKit on Fedora, missing `libicudata.so.66`), in a plain retry loop:

| attempt | leaked drivers (before) | leaked drivers (after) |
|---------|------------------------|------------------------|
| 1       | 1                      | 0                      |
| 2       | 2                      | 0                      |
| 3       | 3                      | 0                      |

- `crawl4ai/browser_manager.py` - Wrap the body of `BrowserManager.start()` in `try/except BaseException` so a failed browser launch rolls back partial startup (stops the leaked Playwright driver process) before re-raising the original error. The diff is mostly indentation — review with `git diff -w`; the real change is 15 added lines.
- `tests/regression/test_reg_browser.py` - Add regression test `test_failed_browser_launch_leaves_no_driver` proving a failed launch propagates the original error, resets manager state, and leaves no driver child process.

## How Has This Been Tested?
- Reproduced the leak first (issue reproducer + a natural failure: WebKit launch on Fedora): drivers grew 1 → 2 → 3 across retries. After the fix: 0 → 0 → 0, with the user-visible error unchanged.
- New regression test verified in both directions: fails on current develop without the fix, passes with it.
- Adversarial matrix (17 tests) covering every `start()` branch: chromium/firefox/persistent-context/CDP launch failures, cancellation mid-start, cleanup that itself raises (original error not masked), retry-then-succeed on the same crawler instance, plus success paths (normal crawl, multi-`arun`, `arun_many`, explicit `start()`/`close()`, sequential crawlers with driver-count baseline).
- Full regression suite: `pytest tests/regression/ -m "not network"` — 293 passed, no new failures.
- Verified `playwright.stop()` does not affect an external CDP browser (stays alive, answers CDP, accepts new clients) — it only ends the driver process we own.

## Checklist:

### DIFF
--- a/crawl4ai/browser_manager.py
+++ b/crawl4ai/browser_manager.py
@@ -626,7 +626,15 @@ class _CDPConnectionCache:
             else:
                 from playwright.async_api import async_playwright
             pw = await async_playwright().start()
-            browser = await pw.chromium.connect_over_cdp(cdp_url)
+            try:
+                browser = await pw.chromium.connect_over_cdp(cdp_url)
+            except BaseException:
+                # Stop the driver we just started so a failed connect doesn't leak it
+                try:
+                    await pw.stop()
+                except BaseException:
+                    pass
+                raise
             cls._cache[cdp_url] = (pw, browser, 1)
             return pw, browser
 
@@ -788,6 +796,25 @@ class BrowserManager:
 
         Note: This method should be called in a separate task to avoid blocking the main event loop.
         """
+        try:
+            await self._start_impl()
+        except BaseException:
+            # Roll back partial startup (e.g. Playwright driver) since a failed __aenter__ never triggers __aexit__
+            try:
+                await self.close()
+            except BaseException:
+                pass
+            if self.playwright is not None:
+                # close() intentionally skips the driver for external-CDP configs; on failed startup we still own it
+                try:
+                    await self.playwright.stop()
+                except BaseException:
+                    pass
+                self.playwright = None
+            self.browser = None
+            raise
+
+    async def _start_impl(self):
         if self.playwright is not None:
             await self.close()
 
--- a/tests/regression/test_reg_browser.py
+++ b/tests/regression/test_reg_browser.py
@@ -7,6 +7,7 @@ stealth mode, session management, network capture, and anti-bot features using
 real browser crawling with no mocking.
 """
 
+import asyncio
 import base64
 import time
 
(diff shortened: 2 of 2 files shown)


ntohidi:
Good root cause write-up, and thanks for the regression test — I ran it both ways, it fails on develop and passes with the fix, which is what I wanted to see.

Three things before this goes in:

**1. `_CDPConnectionCache.acquire()` still leaks.** If `connect_over_cdp` fails at `browser_manager.py:629`, `pw` is already started but never handed back to the manager, so `self.playwright` is still `None` and the rollback skips it. Unreachable CDP endpoint is in your list of covered failures, but that path isn't actually covered. It needs its own try/except inside `acquire()`.

**2. The inner cleanup handlers catch `Exception`, not `BaseException`.** If the task is cancelled again while `close()` is running (normal with `asyncio.timeout` / `wait_for`), `CancelledError` escapes the inner handler, `playwright.stop()` never runs, and the original error gets replaced on the way out. You already picked `BaseException` on the outer catch for exactly this reason — the inner two should match.

**3. Minor:** for `cdp_url` with `cdp_cleanup_on_close=False`, `close()` returns early and leaves `self.browser` set. Worth adding `self.browser = None` to the rollback so state is consistent across all paths.

Separate from those — any chance you could move the body into a `_start_impl()` and make `start()` a thin wrapper? `-w` makes the diff readable, but the re-indent still rewrites blame for the whole function. Not blocking, just kinder to whoever touches it next.

SohamKukreti:
All review points have been addressed. The `_CDPConnectionCache.acquire()` leak has been fixed, inner cleanup handlers now catch `BaseException` as well, `self.browser = None` added to the try catch block so rollback state is consistent and the body has been moved into `_start_impl()` with `start()` as a thin wrapper containing the fix.
Thanks!

ntohidi (approved):
All four points addressed, and the extra test for the cached-CDP path is a nice touch. Ran it locally: both new tests fail on develop and pass here, and the rest of tests/regression is green (the two failures I saw — soft-404 probe and cosine ImportError — reproduce on develop and are unrelated).

Thanks for the _start_impl() split, the diff is much easier to live with now.

PR #2166, opened 2026-08-20
fix: log warning when body visibility check times out silently

### LINKED ISSUE #2144: Body-visibility wait times out silently, making body_visibility_timeout undiscoverable

Battleplus:
## 这次改了什么

When `body_visibility_timeout` expires and `ignore_body_visibility` is `True` (the default), the timeout was completely silent — no log, no error, nothing. Users had no way to discover they needed to adjust `body_visibility_timeout`.

**修法**: Added a debug log message when the timeout fires but is ignored, informing users about the `body_visibility_timeout` parameter.

## 怎么验证的

- Python syntax check passes
- The log message appears in verbose mode when body visibility check times out

## 风险

Extremely low. Only adds a log message on an existing code path. No behavior change.

Closes #2144

### DIFF
--- a/crawl4ai/async_crawler_strategy.py
+++ b/crawl4ai/async_crawler_strategy.py
@@ -826,9 +826,16 @@ class AsyncPlaywrightCrawlerStrategy(AsyncCrawlerStrategy):
                     timeout=30000,
                 )
 
-                if not is_visible and not config.ignore_body_visibility:
+                if not is_visible:
                     visibility_info = await self.check_visibility(page)
-                    raise Error(f"Body element is hidden: {visibility_info}")
+                    if not config.ignore_body_visibility:
+                        raise Error(f"Body element is hidden: {visibility_info}")
+                    else:
+                        self.logger.warning(
+                            message="Body visibility timeout ({elapsed:.1f}s); ignored because ignore_body_visibility=True. Set body_visibility_timeout lower or ignore_body_visibility=False to surface this.",
+                            tag="BODY_VIS",
+                            params={"elapsed": 30.0},
+                        )
 
             except Error:
                 visibility_info = await self.check_visibility(page)


Battleplus:
Superseded by #2179 which skips the entire body visibility check when , eliminating the timeout overhead rather than just logging a warning. Closing this one.

Battleplus:
Superseded by #2179 which skips the entire body visibility check when , eliminating the timeout overhead rather than just logging a warning.

PR #2173, opened 2026-08-21
fix: skip pruning for pre/code tags to preserve whitespace in code blocks

### LINKED ISSUE #2110: PruningContentFilter drops whitespace-only spans in <pre>/<code> — #1181's bug, still present on the fit_markdown path

Battleplus:
Fixes #2110. PruningContentFilter prunes whitespace-only spans inside pre/code tags, corrupting code blocks. Fix: skip pruning for pre and code tags where whitespace is significant.

### DIFF
--- a/crawl4ai/content_filter_strategy.py
+++ b/crawl4ai/content_filter_strategy.py
@@ -712,6 +712,10 @@ class PruningContentFilter(RelevantContentFilter):
         if self._is_preserved(node):
             return
 
+        # Skip pruning for <pre> and <code> tags — whitespace is significant in code blocks
+        if node.name in ("pre", "code"):
+            return
+
         text_len = len(node.get_text(strip=True))
         tag_len = len(node.encode_contents().decode("utf-8"))
         link_text_len = sum(
--- a/tests/async/test_content_filter_prune.py
+++ b/tests/async/test_content_filter_prune.py
@@ -165,6 +165,17 @@ class TestPruningContentFilter:
         second_run = filter.filter_content(basic_html)
         assert first_run == second_run, "Output should be consistent"
 
+    def test_preserves_whitespace_only_nodes_inside_code_blocks(self):
+        html = """<html><body><article><pre><code>if ready:
+<span>    </span>return value</code></pre></article></body></html>"""
+        filter = PruningContentFilter(
+            threshold_type="fixed", threshold=0, min_word_threshold=2
+        )
+
+        contents = filter.filter_content(html)
+
+        assert '<span>    </span>' in "".join(contents)
+
 
 if __name__ == "__main__":
     pytest.main([__file__])


Battleplus:
Added a focused regression test that uses min_word_threshold to reproduce the prior removal of a whitespace-only syntax-highlighting span and asserts the span remains inside the code block. The new test passes. The existing full test_content_filter_prune.py file still has one unrelated baseline failure in test_threshold_types (fixed and dynamic currently both produce one block); the added regression itself is green. Commit: ee66b90.

Battleplus:
Closing as a duplicate after the full audit. Earlier PR #2111 implements the same pre/code subtree guard and already includes substantially broader regression coverage for code whitespace preservation. The focused test added here passed, but consolidating review on #2111 is preferable.

PR #2143, opened 2026-08-16
fix(deep-crawling): allow single-label hostnames in can_process_url validation

Bruce-Yii:
## Summary

Fixes #2079

`BFSDeepCrawlStrategy.can_process_url()` and `BestFirstCrawlingStrategy.can_process_url()` rejected any URL whose netloc contains no dot (`Invalid domain`), silently dropping valid internal targets such as intranet hosts, `localhost` (with or without a port) and Docker service names during deep crawls. The error was raised before the configured `FilterChain` was ever applied.

The scheme (`http`/`https`) and non-empty netloc checks are kept; only the dot heuristic is removed, so `https://name/xyz` now passes validation and reaches the filter chain.

## List of files changed and why

- `crawl4ai/deep_crawling/bfs_strategy.py` - Remove the dot requirement from URL validation in `can_process_url`.
- `crawl4ai/deep_crawling/bff_strategy.py` - Same fix for the best-first strategy.
- `tests/deep_crawling/test_can_process_url.py` - New parametrized tests covering single-label hosts, localhost with port, IP literals, rejection of missing scheme/netloc and non-http(s) schemes, and filter-chain behavior at depth > 0.

## How Has This Been Tested?

- `pytest tests/deep_crawling/test_can_process_url.py -v` → **18 passed**
- `pytest tests/deep_crawling/` → 85 passed; the 2 failures in `test_deep_crawl_resume_integration.py` are pre-existing environment issues (they require launching a real Playwright Chromium, which is not available in this dev venv) and fail identically on unmodified `develop`
- `ruff check --select E9,F63,F7,F82` on the changed files → clean
- `python -m py_compile` on the changed files → clean

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added/updated unit tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2175, opened 2026-08-21
fix: allow single-label hostnames in deep crawl URL validation

### LINKED ISSUE #2079: [Bug]: BFSDeepCrawlStrategy.can_process_url() rejects valid single-label hostnames (netloc without a dot)

Battleplus:
## What

The previous implementation rejected URLs with single-label hostnames (e.g. `https://name/xyz/`) by checking for a dot in netloc. This is overly strict and prevents crawling internal DNS hosts or local development servers.

## Why

Fixes #2079 - BFSDeepCrawlStrategy.can_process_url() rejects valid single-label hostnames.

## Changes

- Removed the dot check in both BFS and BFF deep crawl strategies
- Single-label hostnames are now accepted when the URL has a valid scheme and non-empty netloc

## Testing

- Verified that single-label hostnames are now accepted
- Verified that invalid URLs (missing scheme, invalid scheme) are still rejected

Fixes #2079

### DIFF
--- a/crawl4ai/deep_crawling/bff_strategy.py
+++ b/crawl4ai/deep_crawling/bff_strategy.py
@@ -79,8 +79,8 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
         """
         try:
             parsed = urlparse(url)
-            if not parsed.scheme or not parsed.netloc:
-                raise ValueError("Missing scheme or netloc")
+            if not parsed.scheme or not parsed.hostname:
+                raise ValueError("Missing scheme or hostname")
             if parsed.scheme not in ("http", "https"):
                 raise ValueError("Invalid scheme")
             # Single-label hostnames (e.g. "name" without a dot) are valid
--- a/crawl4ai/deep_crawling/bff_strategy.py
+++ b/crawl4ai/deep_crawling/bff_strategy.py
@@ -83,8 +83,8 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
                 raise ValueError("Missing scheme or netloc")
             if parsed.scheme not in ("http", "https"):
                 raise ValueError("Invalid scheme")
-            if "." not in parsed.netloc:
-                raise ValueError("Invalid domain")
+            # Single-label hostnames (e.g. "name" without a dot) are valid
+            # when the URL has a scheme and non-empty netloc (e.g. internal DNS).
         except Exception as e:
             self.logger.warning(f"Invalid URL: {url}, error: {e}")
             return False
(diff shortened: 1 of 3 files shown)


Battleplus:
Follow-up audit: tightened the relaxed validation to require a parsed hostname (rather than merely a non-empty netloc), so valid single-label/internal hosts and IP literals are accepted while malformed URLs such as http://:8080/path remain rejected. Added parameterized coverage for BFS and Best-First strategies across single-label hosts, IPv4/IPv6, missing schemes, invalid schemes, and missing hostnames. 16 passed. Commit: a515dc1.

Battleplus:
Closing as a duplicate after the full audit. Earlier PR #2143 already covers both BFS and Best-First validation and includes broader targeted coverage for single-label/Docker hosts, localhost, IP literals, malformed netlocs, invalid schemes, and FilterChain execution. I added hostname-hardening and tests here during the audit, but keeping a third PR for the same issue would only split review. Please use #2143.

PR #2187, opened 2026-08-23
fix(overlay): never remove html/body elements during overlay cleanup

### LINKED ISSUE #2161: [Bug]: remove_overlay_elements can remove body if it has a global popup class

Battleplus:
## Summary

`remove_overlay_elements()` removes all elements matching `[class*="popup"]`, `[role="dialog"]`, fixed/absolute positioned elements, etc. If `<body>` has a class containing "popup" (e.g. WordPress theme `qode_popup_menu_push_text_right`), the entire document body gets deleted, leaving an empty page.

## Changes

- `crawl4ai/js_snippet/remove_overlay_elements.js`: Added `isProtected()` guard that skips removal for `<html>` and `<body>` tags in all three removal paths: high-z-index scan, common selector scan, and fixed/sticky position scan.

## Test plan

- [ ] Crawl a site with `<body class="qode_popup_menu_push_text_right">` → body should be preserved
- [ ] Crawl a site with normal popup overlay → overlay should still be removed
- [ ] `remove_overlay_elements=True` with a Cloudflare challenge page → body intact

Fixes #2161

### DIFF
--- a/crawl4ai/js_snippet/remove_overlay_elements.js
+++ b/crawl4ai/js_snippet/remove_overlay_elements.js
@@ -1,4 +1,10 @@
 async () => {
+    // Never remove html or body — they are the document root.
+    const isProtected = (elem) => {
+        const tag = elem.tagName;
+        return tag === 'HTML' || tag === 'BODY';
+    };
+
     // Function to check if element is visible
     const isVisible = (elem) => {
         const style = window.getComputedStyle(elem);
@@ -59,6 +65,7 @@ async () => {
             const position = style.position;
 
             if (
+                !isProtected(elem) &&
                 isVisible(elem) &&
                 (zIndex > 999 || position === "fixed" || position === "absolute") &&
                 (elem.offsetWidth > window.innerWidth * 0.5 ||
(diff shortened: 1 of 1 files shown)


SohamKukreti:
Please avoid opening duplicate PRs on issues someone else is working on

PR #2212, opened 2026-08-28
Make the untrusted timeout ceiling configurable

### LINKED ISSUE #2211: page_timeout is silently capped at 60s over HTTP with no way to raise it
## Problem

`_clamp_untrusted` caps `page_timeout` and `wait_for_timeout` at `_MAX_TIMEOUT_MS = 60_000` for any config arriving over HTTP. The Docker server passes `Provenance.UNTRUSTED` at every `/crawl` call site, so a client asking for more is silently given 60s — no warning, no field in the response. A page that legitimately takes longer always fails with `Page.goto: Timeout 60000ms exceeded`, quoting a number the caller never sent.

The cap itself is reasonable for a public server. Not being able to change it is the problem: it isn't in `config.yml`, no env var overrides it, and the value is a module-level literal.

`crawler.base_config` looks like the intended escape hatch and isn't — it's applied post-deserialization with `setattr`, genuinely bypassing the clamp, but only for fields that are `None` or `''`. `page_timeout` always carries its 60000 default, so setting it there does nothing, silently.

## Expected

An operator running their own container can raise the ceiling — an env var, a `limits:` entry alongside `wall_clock_s`, or a `base_config` that applies to defaults too. Failing that, a clamped request should say so rather than report a timeout the caller didn't ask for.

## Reproduce (docker, 0.9.2)

A server that answers after 90s:

```python
import time
BODY = b"<html><body><h1>Slow</h1><p>" + b"x"*600 + b"</p></body></html>"
class H(BaseHTTPRequestHandler):
    def do_GET(self):
        time.sleep(90)
        self.send_response(200); self.send_header("Content-Length", str(len(BODY))); self.end_headers()
        self.wfile.write(BODY)
HTTPServer(("0.0.0.0", 8080), H).serve_forever()
```

```bash
docker network create repro
docker run -d --name slowsrv --network repro -v $PWD/slow_server.py:/srv/s.py:ro python:3.12-slim python /srv/s.py
docker run -d --name c4ai --network repro -p 11235:11235 --shm-size=2gb \
  -e CRAWL4AI_API_TOKEN=dev-token -e CRAWL4AI_ALLOW_INTERNAL_URLS=true unclecode/crawl4ai:0.9.2

curl -X POST http://localhost:11235/crawl -H 'authorization: Bearer dev-token' \
  -H 'content-type: application/json' \
  -d '{"urls":["http://slowsrv:8080/"],"crawler_config":{"page_timeout":300000,"wait_until":"domcontentloaded"}}'
```

**Before** — asked for 300s:

```
HTTP 500 in 60.7s
Page.goto: Timeout 60000ms exceeded
```

**After** — same request, against an image built with `_MAX_TIMEOUT_MS = 600_000`:

```
HTTP 200 in 90.9s
success: true, markdown: "# Slow page\nxxxx…"
```

## Suggested patch

Make the ceiling configurable rather than removing it:

```python
_MAX_TIMEOUT_MS = int(os.environ.get("CRAWL4AI_MAX_TIMEOUT_MS", 60_000))
```

An env var keeps the safe default for a public deployment while letting an operator raise it for a trusted one. A `limits.max_timeout_ms` key in `config.yml`, next to `wall_clock_s`, would fit the existing "0 = unbounded" convention equally well.

Separately, `base_config` skipping any field with a non-empty default is surprising enough to be worth documenting.

damusix:
Closes #2211.

`_clamp_untrusted` caps `page_timeout`, `wait_for_timeout` and `body_visibility_timeout` at 60s for any config arriving over HTTP. The bound is right for a server reachable by untrusted callers — the problem is that it was a module literal with no env var, no `config.yml` key, and no way for an operator to change it, so a private deployment could not crawl a page that legitimately takes longer. The failure is also confusing: `Page.goto: Timeout 60000ms exceeded` quotes a number the caller never sent.

`CRAWL4AI_MAX_TIMEOUT_MS` now sets the ceiling and defaults to the same 60000ms, so nothing changes for anyone who does not set it. A smaller value tightens the bound; a request still only ever gets the timeout it asks for.

A value that is not a positive integer warns and keeps the default rather than being coerced. A typo in this particular variable would silently widen a DoS bound, which seemed worth being loud about.

The lookup happens per call rather than being captured at import, so the setting applies wherever the process picked its environment up, and a test can set it without reloading the module. `_MAX_TIMEOUT_MS` is renamed to `_DEFAULT_MAX_TIMEOUT_MS` to stop the name claiming something it no longer is; it is private and had no references outside this module.

**Verified**

Against `unclecode/crawl4ai:0.9.2` in docker, crawling a server that answers after 90s with `page_timeout: 300000`:

| | result |
| --- | --- |
| stock | `HTTP 500` in 60.7s — `Page.goto: Timeout 60000ms exceeded` |
| ceiling raised | `HTTP 200` in 90.9s, markdown returned |

15 new tests in `tests/test_config_defaults.py` cover the default, raising, tightening, a request above a raised ceiling, every invalid value, the warning, and that trusted configs stay unclamped. `tests/test_config_defaults.py` passes (58). Reverting the source change fails 8 of them.

`black --diff` leaves the added lines unchanged; the existing reformat suggestions in both files are pre-existing and left alone.

### DIFF
--- a/crawl4ai/async_configs.py
+++ b/crawl4ai/async_configs.py
@@ -270,9 +270,44 @@ UNTRUSTED_FIELD_ALLOWLIST = {
 }
 
 # Upper bounds applied to attacker-influenced quantities after filtering.
-_MAX_TIMEOUT_MS = 60_000
+_DEFAULT_MAX_TIMEOUT_MS = 60_000
 _MAX_SCROLL_STEPS = 1000
 _MAX_VIEWPORT = 4000
+
+
+def _max_timeout_ms() -> int:
+    """Ceiling for the untrusted timeout fields, in milliseconds.
+
+    60s is the right bound for a server reachable by untrusted callers, and
+    stays the default. An operator whose deployment is not public — a crawler
+    on a private network fetching pages that legitimately take minutes — can
+    raise it with CRAWL4AI_MAX_TIMEOUT_MS, or lower it to tighten the bound.
+
+    Read per call rather than captured at import so the setting applies
+    wherever the process picked its environment up, and so a test can set it
+    without reloading the module. A value that is not a positive integer is
+    refused loudly and the default kept: a typo here would silently widen a
+    DoS bound, which is the one outcome worse than the timeout being fixed.
+    """
+    raw = os.getenv("CRAWL4AI_MAX_TIMEOUT_MS")
+
+    if raw is None or raw == "":
+        return _DEFAULT_MAX_TIMEOUT_MS
+
+    try:
+        ceiling = int(raw)
+    except ValueError:
+        ceiling = 0
+
+    if ceiling <= 0:
+        warnings.warn(
+            f"CRAWL4AI_MAX_TIMEOUT_MS={raw!r} is not a positive integer; "
+            f"keeping the {_DEFAULT_MAX_TIMEOUT_MS}ms default.",
+            stacklevel=2,
+        )
+        return _DEFAULT_MAX_TIMEOUT_MS
+
+    return ceiling
 
 
 def _filter_untrusted_fields(type_name: str, params: dict) -> dict:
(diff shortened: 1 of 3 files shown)


damusix:
@unclecode

Issue #2205, opened 2026-08-27
[Bug]: Post-navigation page operations have no timeout, and page cleanup is skipped on task cancellation — pages can leak and crawls can hang forever


SohamKukreti:
## Summary

Found while investigating #2202 (Docker containers accumulating renderer processes for weeks). Two related defects in `async_crawler_strategy.py` make it possible for a crawl to hang indefinitely and for its page to never be closed.

### 1. Nothing after navigation has a timeout

`page_timeout` only bounds `page.goto()` (`async_crawler_strategy.py:762-764`). Every page interaction after navigation is an un-timed call into the page's JS engine:

- overlay/consent removal — `remove_overlay_elements` (`:1550`)
- body-visibility check, `css_selector` extraction (`:1102`), image-dimension updates
- `page.content()` (`:1115`)
- shadow-DOM flattening, iframe processing, user `js_code`

The adapter is a plain pass-through to `page.evaluate` (`browser_adapter.py:61-65`), which has no timeout in Playwright. The repo never calls `set_default_timeout` except inside `if self.config.accept_downloads:` (`browser_manager.py:1215-1217`) — and default timeouts would not cover `evaluate` anyway. Only `scan_full_page` is wrapped in `asyncio.wait_for`.

A page whose main thread goes busy after `DOMContentLoaded` (bad loop, broken ad script, hostile page) therefore hangs the crawl **forever**, holding its page (= one renderer process) open. Verified with a page that starts a busy loop right after DCL: navigation succeeds, then `arun()` never returns, far past `page_timeout`.

### 2. The page-closing `finally` doesn't survive cancellation

The cleanup block (`async_crawler_strategy.py:1216-1231`) guards with `except Exception`. `asyncio.CancelledError` is a `BaseException`, so a task cancelled during the first `await` in that block (`release_page_with_context`) propagates out and `page.close()` is never reached — the page and its context refcount leak. Reachable from dispatcher/stream teardown paths that cancel in-flight tasks.

## Impact

In long-lived processes (the Docker server's browser pool, any SDK user reusing a crawler), leaked pages accumulate indefinitely. Combined with the launch flags that disable background throttling, each leaked page also burns CPU forever. This is the core mechanism behind the multi-week renderer accumulation in #2202.

## Proposed fix

- Wrap the post-navigation phase (or at minimum every `evaluate`/`content()` call) in `asyncio.wait_for` with a budget derived from `page_timeout`, so a crawl always terminates.
[...]

Issue #2204, opened 2026-08-27
[Bug]: Docker pool's permanent browser never serves any request - signature is computed without the egress proxy, so no `/crawl` can ever match it


SohamKukreti:
## Summary

Found while investigating #2202. The Docker server's "permanent" warm browser is dead weight: it is started at boot, serves zero requests, is never cleaned up, and causes every container to run a second browser tree from the first request onward.

## Root cause

The pool matches requests to browsers by hashing the full `BrowserConfig` (`crawler_pool.py:46-49`), and `proxy_config` is part of `to_dict()` (`async_configs.py:968`).

- `init_permanent()` is called at startup with a config built straight from `config.yml` — **without** `enforce_egress()` (`server.py:199-203`), so its fingerprint has `proxy_config: None`.
- Every `/crawl` request's config goes through `enforce_egress()` first (`api.py:687-690`), which sets `proxy_config` to the egress pinning proxy — whose port is random each boot (`egress_broker.py:198-201`).

The two fingerprints can never be equal, so `_is_default_config()` (`crawler_pool.py:60`) never matches and `PERMANENT` is never returned. Note `get_default_browser_config()` (`server.py:126-141`) *does* apply `enforce_egress`, so even the server's own `/html`, `/screenshot`, `/pdf`, `/execute_js` endpoints miss it.

Additionally, the janitor only sweeps `HOT_POOL` and `COLD_POOL` (`crawler_pool.py:177,198`) — `PERMANENT` is never inspected, so the unused browser also can never be reclaimed.

## Impact

- ~270 MB RSS + one Playwright driver + full Chromium tree per container, doing nothing, forever.
- Every container shows two `driver → browser` process branches after the first request (observed in #2202's process listing: one branch from boot, one from the first crawl a day later).
- The `text_mode` optimization configured for the default browser never applies to real traffic.

## Verified

Reproduced on `unclecode/crawl4ai:0.9.2`: fresh container = one browser tree; after a single plain `/crawl` = two independent trees, permanent one idle.

## Proposed fix

Keep the permanent browser, fix the match: build its config through the same path requests use (`get_default_browser_config()`, i.e. **after** `enforce_egress`), and compute `DEFAULT_CONFIG_SIG` from that. Alternatively/additionally, exclude the server-injected `proxy_config` from the pool signature, since the server sets it identically on every request.

SohamKukreti:
I'll be picking this one up

PR #2214, opened 2026-08-29
Make the browser pool un-pinnable by hung crawls

### LINKED ISSUE #2202: [Bug]: Browser pages and renderer processes accumulate for weeks on non-streaming /crawl (Docker 0.9.2)
### crawl4ai version

0.9.2 (Docker image `unclecode/crawl4ai:0.9.2`, `C4AI_VERSION=0.9.2`)

### Expected Behavior

After a non-streaming `POST /crawl` completes, the pages and browser contexts it opened are closed. A container that has served no traffic for two weeks sits at near-zero CPU and holds only the warm browser pool.

### Current Behavior

The container had been up 29 days serving a low volume of non-streaming `/crawl` requests (a handful per day, `max_depth=2`, at most 5 pages). At inspection it held 46 chromium processes and had not served a request for 15 days:

```
 3  chrome-headless-shell (browser root)
 2  playwright/driver/node
35  chrome-headless-shell --type=renderer
 4  --type=zygote
 3  --type=utility
 2  --type=gpu
```

945 MB RSS across those processes, 957 MiB for the container, and a steady **31% of a CPU core with zero traffic** — two consecutive `docker stats` samples read 30.69% and 30.86%.

Renderer start timestamps cluster on the days crawls actually ran: dozens on Jul 29 between 19:41 and 23:21, isolated ones later, the newest on Aug 11. A `--type=renderer` process lives exactly as long as its page, so those pages were never closed and the renderers outlived their requests by weeks.

There are also two independent `driver → browser root` branches, started Jul 28 14:25:55 and Jul 29 17:21:15. The older branch was never torn down when the newer one started.

Restarting the container dropped it to 7 chromium processes, 426 MiB and 0.4% CPU.

### Is this reproducible?

Yes

### Inputs Causing the Bug

Plain non-streaming request — no session id, no hooks:

```json
{
  "urls": ["<single http(s) url>"],
  "browser_config": {"type": "BrowserConfig", "params": {"headless": true, "accept_downloads": false}},
  "crawler_config": {"type": "CrawlerRunConfig", "params": {
    "stream": false,
    "cache_mode": "bypass",
    "wait_until": "domcontentloaded",
    "page_timeout": 30000,
    "process_iframes": false,
    "remove_overlay_elements": true,
    "exclude_external_images": true
  }}
}
```

### Steps to Reproduce

1. Run `unclecode/crawl4ai:0.9.2` in Docker with no memory or CPU limit on a cgroup v2 host.
2. Send the request above a few times a day for a few days, then stop sending traffic.
3. After some days, inspect the process table inside the container.

### OS

Linux, Docker, cgroup v2 host

### Python version

3.12.13 (image default)

### Browser

chromium_headless_shell-1228, bundled with the image

### Error logs & Screenshots

Nothing is logged and the container reports `healthy` the whole time.

Possibly related: #943 (closed, fixed in 0.8.5), #2083, #2155, #2123. This container had no memory limit, so per #2123 the memory guard reads host RAM instead of container usage — that may be why nothing reclaims these pages.

SohamKukreti:
## Summary

Fixes #2202

A crawl that hangs inside the browser (e.g. a page whose JS thread never yields — see #2205)
used to pin its pooled browser forever: the busy counter never decremented, the janitor
refused to touch a "busy" browser unconditionally, and the per-crawl deadline shipped
disabled. Weeks of light traffic accumulated dozens of chromium processes and ~1 GB RSS
(reporter's container: 46 chrome processes, 35 never-closed pages).

This PR makes the server survive hung or cancelled crawls without leaking browsers:

- **Cancellation-safe release**: `release_crawler` no longer takes the pool LOCK and has no
  await points, so a request cancelled mid-`finally` (client disconnect, deadline) can never
  skip the busy-counter decrement.
- **Janitor stale-lease backstop**: a browser reporting "busy" that nothing has touched for
  longer than `crawler.pool.stale_lease_s` (default `0` = auto `max(2 × wall_clock_s, 21600)`;
  6 h floor because streaming has no deadline) is treated as pinned by a leaked request and
  force-closed with an ERROR log naming the signature.
- **Closes never block the pool**: the janitor and `close_all()` route browser closes through
  a fire-and-forget background task (60 s cap, logged on timeout) instead of awaiting
  `close()` while holding the pool LOCK — a wedged browser previously froze `get_crawler()`
  and all future janitor passes for the whole server. `close_all()` drains pending closes
  (65 s cap) so shutdown is clean.
- **Per-crawl deadline on by default**: `limits.wall_clock_s` is now `1800` (was `0`). A
  non-streaming crawl exceeding it returns 504 and releases its browser. `0` remains the
  documented opt-out; the backstop above covers that case.

Out of scope (recorded for follow-up): fixing *why* crawls hang (#2205), the permanent
browser fingerprint (#2204), deadlines for the streaming path and the other pooled
endpoints, and `monitor_routes.py`'s closes-under-LOCK.

- `deploy/docker/crawler_pool.py` — the fix: lock-free `release_crawler`, stale-lease
  backstop in both janitor sweeps, `_close_in_background` helper, `close_all` drain,
  ceiling config with safe fallback (bad YAML values, including `true`, fall back to auto).
- `deploy/docker/config.yml` — `wall_clock_s: 1800` default; new `pool.stale_lease_s` knob.
- `deploy/docker/MIGRATION.md` — documents the new 504 default and the janitor force-close
  behavior, including the "raise `stale_lease_s` if your crawls legitimately run > 6 h" note.
- `deploy/docker/ARCHITECTURE.md` — janitor pseudo-code and pool config reference updated to
  match the new design.
- `deploy/docker/utils.py` — `dns.resolver` import deferred into `verify_email_domain()` so
  the module imports without dnspython (needed for the unit tests; dnspython ships in the
  image, so no runtime change).
- `tests/docker/test_pool_release.py` — rewritten to exercise the real `crawler_pool` module
  instead of a standalone copy of the logic; covers release/cancellation safety, both sweep
  paths, the backstop, the ceiling config resolution, and shutdown draining.

## How Has This Been Tested?

- **Unit**: 28 tests against the real module (no Docker needed):
  `python -m pytest tests/docker/test_pool_release.py` — release decrements without the LOCK
  and survives cancellation; janitor force-closes busy-past-ceiling browsers (cold and hot),
  skips genuinely busy ones, and is not blocked by a `close()` that hangs; `close_all` waits
  for pending background closes and never holds the LOCK while one hangs; ceiling resolution
  rejects nonsense values (`-1`, `"30m"`, `true`, `null`).
- **Live acceptance (scripted)**: full suite against a locally built image plus a control run
  on the unfixed image — 14/14 pass. Hung crawl 504s within deadline+10 s and the janitor
  reaps the browser (renderers back to baseline); 20 client cancellations and 40 concurrent
  hung crawls never wedge the server; 60 healthy crawls leak nothing and trigger zero
  force-closes; a genuinely busy browser is never touched; with `wall_clock_s: 0` the
  backstop force-closes the pinned browser, kills the wedged renderer, and the server keeps
  serving; container shutdown exits 0 with zero destroyed-task warnings. The unfixed image
  reproduces the original bug (pinned browser never reclaimed).
- Both shutdown orderings verified live: stop before the janitor reaps (close_all closes a
  live pooled browser via the background path) and stop after a background reap.

## Checklist:

### DIFF
--- a/deploy/docker/ARCHITECTURE.md
+++ b/deploy/docker/ARCHITECTURE.md
@@ -272,20 +272,30 @@ async def janitor():
         await asyncio.sleep(interval)
 
         async with LOCK:
-            # Clean cold pool first (less valuable)
-            for sig in list(COLD_POOL.keys()):
-                if now - LAST_USED[sig] > cold_ttl:
-                    await COLD_POOL[sig].close()
-                    del COLD_POOL[sig], LAST_USED[sig], USAGE_COUNT[sig]
-                    await track_janitor_event("close_cold", sig, {...})
-
-            # Clean hot pool (more conservative)
-            for sig in list(HOT_POOL.keys()):
-                if now - LAST_USED[sig] > hot_ttl:
-                    await HOT_POOL[sig].close()
-                    del HOT_POOL[sig], LAST_USED[sig], USAGE_COUNT[sig]
-                    await track_janitor_event("close_hot", sig, {...})
+            # Both pools, same rule (cold first; hot gets the longer TTL):
+            for sig in list(pool.keys()):
+                idle = now - LAST_USED[sig]
+                if idle <= ttl:
+                    continue                      # used recently
+                if pool[sig].active_requests > 0:
+                    if idle <= STALE_CEILING:
+                        continue                  # genuinely serving a request
+                    log.error("Leaked request counter ... force-closing")
+                    # busy past the ceiling = a hung/leaked request pinned it
+                _close_in_background(pool[sig])   # close() can hang; NEVER awaited under LOCK
+                del pool[sig], LAST_USED[sig], USAGE_COUNT[sig]
+                await track_janitor_event("close_cold"/"close_hot", sig, {...})
 ```
+
+Two safeguards keep the pool un-pinnable (issue #2202):
+
+- **Stale-lease backstop**: `active_requests` is only trusted while the browser was
+  touched within `STALE_CEILING` (`pool.stale_lease_s`; `0` = auto
+  `max(2 × limits.wall_clock_s, 21600)`). A counter stuck past that means a request
+  hung or leaked, so the browser is force-closed instead of pinned forever.
+- **Background close**: `close()` on a wedged browser can hang, so the janitor never
+  awaits it while holding LOCK. Closes run as fire-and-forget tasks with a 60s cap;
+  `close_all()` drains them (65s cap) so shutdown is clean.
 
 **Config Signature Generation:**
 
@@ -812,6 +822,7 @@ crawler:
 
   pool:
     idle_ttl_sec: 300  # Base TTL for cold pool (5 min)
+    stale_lease_s: 0   # Leak-backstop ceiling; 0 = auto: max(2 x limits.wall_clock_s, 21600)
 
   rate_limiter:
     enabled: true
--- a/deploy/docker/MIGRATION.md
+++ b/deploy/docker/MIGRATION.md
@@ -161,7 +161,7 @@ longer published. For an **external** redis, set `REDIS_PASSWORD`.
 ```yaml
 limits:
   max_body_bytes: 10485760   # request body cap (413); 0 = unbounded
-  wall_clock_s: 0            # per-crawl deadline (504); 0 = none
+  wall_clock_s: 1800         # per-crawl deadline (504); 0 = none
   queue:
     maxsize: 1000            # background job queue (503 when full); 0 = unbounded
     workers: 4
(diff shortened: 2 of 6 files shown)


ntohidi:
For my own reference: 

The code is fine. The defaults are the risk. It changes behavior for existing users.  The real risk: wall_clock_s goes from 0 (off) to 1800 (30 min).

Anyone whose batch or deep crawl takes over 30 minutes today gets a 504 tomorrow. And they get nothing back. the deadline wraps the whole arun_many() call, so no partial results. It's in MIGRATION.md, but people upgrade without reading.
  
The reporter's own case was max_depth=2, 5 pages. Fine. A 200-URL batch is not.
  
Smaller risk: the janitor can kill a live browser if a crawl runs past 6 hours. Only bites streaming users with very long crawls. The 6h floor and the warning in the docs make this unlikely.

PR #2224, opened 2026-09-01
Skip /config/dump pre-flight for md/llm endpoints in playground

### LINKED ISSUE #2222: [Bug]: Playground: /md and /llm endpoints always fail because Advanced Config `code` is sent to /config/dump
### crawl4ai version

v0.9.3

### Expected Behavior

The Playground (/playground/) should work for all endpoint modes. The Advanced Config editor is a /crawl-only feature (the panel is hidden for md/llm), so selecting the md endpoint and clicking Run should simply POST { url, f, q, c } to /md — the server builds its own CrawlerRunConfig internally (handle_markdown_request). No client-side config validation should be involved.

### Current Behavior

**runCrawl()** in **deploy/docker/static/playground/index.html** calls **pyConfigToJson()** unconditionally for every endpoint, including md and llm. pyConfigToJson() reads the Advanced Config CodeMirror editor (which defaults to the Python snippet CrawlerRunConfig(stream=True, cache_mode=CacheMode.BYPASS,)) and POSTs it to /config/dump using the removed legacy `code` protocol:

```json
{ "type": "CrawlerRunConfig", "code": "CrawlerRunConfig(\n    stream=True,\n    cache_mode=CacheMode.BYPASS,\n)" }
```

Since the security hardening (eval removal from /config/dump + untrusted trust boundary), the server parses request bodies with Provenance.UNTRUSTED and code is a globally forbidden field, so the request is rejected with:

```json
{ "detail": "field 'code' is not permitted on CrawlerRunConfig from an untrusted request" }
```

In runCrawl() the 400 error triggers the catch branch. The fallback (minimal { crawler_config: { type: 'CrawlerRunConfig', params: { stream: true } } }) only applies when endpoint === 'crawl' and the editor text contains stream=True. For md/llm the code path aborts the whole run, so the md mode is unusable even without touching the Advanced Config. The only workaround today is pasting empty text / relying on nothing, but since the editor has a non-empty default template, pyConfigToJson() always runs.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- URL(s): https://example.com
- Endpoint: md (and llm)
- Advanced Config: default
```

### Steps to Reproduce

```bash
1. Start the Docker server and open http://localhost:11235/playground/
2. Leave the endpoint set to "md" (default), enter any URL, leave Advanced Config untouched
3. Click "Run"
4. Observe: status flips to "error", response shows
`{ "error": "field 'code' is not permitted on CrawlerRunConfig from an untrusted request" }`
and the /md request is never sent
```

### Code snippets

```python

```

### OS

Docker

### Python version

3.12.14

### Browser

Chrome

### Browser version

_No response_

Image

SohamKukreti:
## Summary
Fixes #2222

The playground validated the Advanced Config editor via `/config/dump` for every endpoint, including `md` and `llm` where the panel is hidden and its result is never used. That pre-flight sends the legacy `{ type, code }` protocol, which the server removed in the 0.8.x security fixes and rejects loudly since 0.9.3 (`code` is a forbidden field on untrusted requests). The fallback only rescued `crawl`, so `md`/`llm` (and `crawl_stream`) aborted with `field 'code' is not permitted on CrawlerRunConfig from an untrusted request` before their request was ever sent.

The fix skips the `/config/dump` pre-flight unless the endpoint is `crawl` or `crawl_stream`. Behavior inside the guard is unchanged.

- `deploy/docker/static/playground/index.html` - wrap the `/config/dump` pre-flight in `runCrawl()` with an endpoint guard so `md`/`llm` skip it; drop the now-redundant `isCrawlEndpoint` check inside the catch fallback.

## How Has This Been Tested?
Manually verified and tested the playground
and
Drove the real playground UI with headless Chromium against a clean `unclecode/crawl4ai:0.9.3` Docker container (the affected release):

- `md`: fit / raw / bm25+query / llm-filter, both cache modes - all succeed; only `POST /md` fires, no `/config/dump` call
- `llm` with a question - succeeds via `GET /llm/{url}?q=...`
- `crawl` with the default snippet - behavior unchanged: `/config/dump` -> fallback -> `/crawl`, succeeds
- `crawl_stream` with the default snippet - now works (streams from `/crawl/stream`; on stock 0.9.3 it aborted like `md`)
- Error paths intact: an invalid URL surfaces the server error; editor content can no longer affect `md`/`llm`

## Checklist:

### DIFF
--- a/deploy/docker/static/playground/index.html
+++ b/deploy/docker/static/playground/index.html
@@ -686,29 +686,31 @@
         async function runCrawl() {
             const endpoint = document.getElementById('endpoint').value;
             const urls = document.getElementById('urls').value.trim().split(/\n/).filter(u => u);
-            // 1) grab python from CodeMirror, validate via /config/dump
+            // 1) grab python from CodeMirror, validate via /config/dump.
+            // Advanced Config is /crawl-only, so md/llm skip this entirely.
             let advConfig = {};
-            try {
-                const cfgJson = await pyConfigToJson(); // may throw
-                if (Object.keys(cfgJson).length) {
-                    const cfgType = document.getElementById('cfg-type').value;
-                    advConfig = cfgType === 'CrawlerRunConfig'
-                        ? { crawler_config: cfgJson }
-                        : { browser_config: cfgJson };
-                }
-            } catch (err) {
-                const codeText = cm.getValue();
-                const streamFlag = /stream\s*=\s*True/i.test(codeText);
-                const isCrawlEndpoint = document.getElementById('endpoint').value === 'crawl';
-                if (isCrawlEndpoint && streamFlag) {
-                    // Fallback: minimal dump-shaped config so shouldUseStream + server load agree
-                    advConfig = { crawler_config: { type: 'CrawlerRunConfig', params: { stream: true } } };
-                } else {
-                    updateStatus('error');
-                    document.querySelector('#response-content code').textContent =
-                        JSON.stringify({ error: err.message }, null, 2);
-                    forceHighlightElement(document.querySelector('#response-content code'));
-                    return; // stop run
+            if (endpoint === 'crawl' || endpoint === 'crawl_stream') {
+                try {
+                    const cfgJson = await pyConfigToJson(); // may throw
+                    if (Object.keys(cfgJson).length) {
+                        const cfgType = document.getElementById('cfg-type').value;
+                        advConfig = cfgType === 'CrawlerRunConfig'
+                            ? { crawler_config: cfgJson }
+                            : { browser_config: cfgJson };
+                    }
+                } catch (err) {
+                    const codeText = cm.getValue();
+                    const streamFlag = /stream\s*=\s*True/i.test(codeText);
+                    if (streamFlag) {
+                        // Fallback: minimal dump-shaped config so shouldUseStream + server load agree
+                        advConfig = { crawler_config: { type: 'CrawlerRunConfig', params: { stream: true } } };
+                    } else {
+                        updateStatus('error');
+                        document.querySelector('#response-content code').textContent =
+                            JSON.stringify({ error: err.message }, null, 2);
+                        forceHighlightElement(document.querySelector('#response-content code'));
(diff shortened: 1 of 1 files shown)


ntohidi (approved):
Diagnosis is right and I'd take this — `md`/`llm` were failing on a pre-flight whose result they never used, and dropping the `isCrawlEndpoint` check is what actually unblocks `crawl_stream`. Both good.

But this fixes half of #2222, and the other half is worth saying out loud before the issue gets closed.

`/config/dump` takes `{type, params}` now. The playground still sends `{type, code}`, and `code` is on the untrusted forbidden list, so that call **always** 400s — for `crawl` too, not just `md`/`llm`. What's left on `crawl` is the regex fallback, which means:

- Default snippet (`stream=True, cache_mode=CacheMode.BYPASS`) → fallback fires, request goes out as `{stream: true}`. The `cache_mode` line is silently dropped. Looks like it worked.
- Switch type to BrowserConfig → that template has no `stream=True`, so no fallback, hard `✖ config error` and the run aborts. The panel is unusable.

So the Advanced Config editor is dead on the one endpoint that still shows it, and on the default path it quietly discards whatever the user typed beyond `stream`. That's arguably worse than the loud error `md` was giving.

Not asking you to fix it here — the real fix isn't a playground tweak. The server dropped `code` on purpose (it was a construction oracle), so a Python-snippet editor can't come back. It needs either a params/JSON editor or the panel removed, and that's a separate PR with a UI decision in it.

Two small things in this one, fine as follow-ups:

- `pyConfigToJson()` is what writes `#cfg-status`. `md`/`llm` no longer call it, so a `✖ config error` left over from a previous `crawl` run stays on screen.
- The fallback comment says "so shouldUseStream + server load agree", which reads like the fallback is a normal mode. It's a workaround for a broken pre-flight. A line saying so would save the next person some digging.

Merging this as the one-line unblock. Will link a follow-up issue for the editor so it doesn't get closed with #2222.

ntohidi:
Follow-up filed as #2260 for the Advanced Config panel.

PR #2229, opened 2026-09-05
fix: preventing disallow: /*? from blocking the whole website.

### LINKED ISSUE #2225: [Bug]: Disallow: /*? in robots.txt blocks the whole website.
### crawl4ai version

0.9.0

### Expected Behavior

The rule in robots.txt disallows scraping urls with query params, the rest of the page works as expected.

### Current Behavior

crawl4ai 0.9.x patches the stdlib urllib.robotparser to support wildcards, but it still matches rules against normalized paths (paths where the query string has already been stripped). Ecommerce sites commonly ship (this is advised by multiple SEO blogs online):

`Disallow: /*?`

The intent is to block only URLs carrying a query string, e.g. /?s=search. After normalization the ? is gone, so the rule collapses to Disallow: /* and the entire site reads as disallowed.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- Disallow: /*? in robots.txt
```
When check_robots_txt=True

### Steps to Reproduce

```bash
- Scrape any website (check_robots_txt=True) with Disallow: /*? in robots.txt (test webserver is sufficient in this case)
```

### Code snippets

```python

```

### OS

macOS

### Python version

3.9.10

### Browser

Chrome

### Browser version

_No response_

_No response_

Nalhin:
## Summary

Fixes #2225 - disallow: /*? in robots.txt blocks the whole website.

Adding logic in robots parser to append "*" after "?" so that urls ending with "?" get parsed correctly by the lib.

## How Has This Been Tested?
Unit tests for this specific edge case.

## Checklist:

### DIFF
--- a/crawl4ai/utils.py
+++ b/crawl4ai/utils.py
@@ -249,6 +249,22 @@ class VersionManager:
         return installed is None or installed < current
 
 
+def _preserve_bare_query(rules_text: str) -> str:
+    """Append '*' to Allow/Disallow values ending in a bare '?' (e.g. '/*?').
+
+    '/*?' and '/*?*' allow/deny exactly the same URLs; the rewrite only
+    survives parsers that would otherwise drop the trailing '?'.
+    """
+    fixed = []
+    for raw_line in rules_text.splitlines():
+        body, _, _ = raw_line.partition("#")
+        key, sep, value = body.partition(":")
+        if sep and key.strip().lower() in ("allow", "disallow") and value.strip().endswith("?"):
+            raw_line = f"{key.strip()}: {value.strip()}*"
+        fixed.append(raw_line)
+    return "\n".join(fixed)
+
+
 class RobotsParser:
     # Default 7 days cache TTL
     CACHE_TTL = 7 * 24 * 60 * 60
@@ -355,8 +371,10 @@ class RobotsParser:
             return True
 
         # Create parser for this check
-        parser = RobotFileParser() 
-        parser.parse(rules.splitlines())
+        parser = RobotFileParser()
+        # Old Pythons drop a trailing '?' from rules, so '/*?' becomes
+        # '/*' and blocks the whole site. '/*?*' matches the same URLs.
+        parser.parse(_preserve_bare_query(rules).splitlines())
         
         # If parser can't read rules, allow access
         if not parser.mtime():
--- a/tests/general/test_robot_parser.py
+++ b/tests/general/test_robot_parser.py
@@ -123,6 +123,44 @@ Allow: /public/
         finally:
             await runner.cleanup()
 
+        # 4b. Test query-string disallow (Disallow: /*?) on a separate host port.
+        # Plain URLs must stay crawlable while query URLs are denied (RFC 9309).
+        async def start_query_server():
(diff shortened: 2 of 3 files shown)


SohamKukreti:
Verified locally and this looks good to merge.

**What I checked**

- Cherry-picked the commit onto current `develop` (clean).
- Confirmed the root cause: on Python 3.9–3.12, `RuleLine.__init__` does `urlunparse(urlparse(path))`, which drops a trailing empty `?`. `/*?` is stored as `/%2A`, and the wildcard patch then turns that into `^/.*`. Python 3.13 switched to `normalize_path()`, which keeps the `?`, so the bug only affects 3.9–3.12.
- `tests/unit/test_robots_query_rules.py`: 22 passed on 3.10 and 3.13 with the fix. Without the fix: 10 failed on 3.10 (the 7 helper tests plus the 3 plain-URL tests), 7 failed on 3.13 (helper tests only, as expected since 3.13 has no bug).
- Live check against `https://www.wired.com/robots.txt` (`User-agent: *` / `Disallow: /*?`) on Python 3.10: before the fix, `arun("https://www.wired.com/", check_robots_txt=True)` returned 403 "Access denied by robots.txt"; after the fix it crawls, and `/search?q=ai` is still denied.
- Rewrite is semantically a no-op: robots rules are prefix matches, so `/x?` and `/x?*` match the same set, and rules ending in `$` are never touched. Ran the helper over six other live robots files; only lines ending in a bare `?` change.
- Full `tests/unit` suite: 102 passed on 3.10.

**Two small nits (non-blocking)**

1. `test_preserve_bare_query_keeps_document_structure` asserts that the trailing newline is dropped. That pins an incidental side effect of `"\n".join(...)` rather than intended behavior; if the helper is ever changed to preserve the newline, the test would fail for no real reason. Suggest dropping the first assert and keeping only the `.splitlines()` comparison.

2. Part 4b in `tests/general/test_robot_parser.py` binds a second server to a hardcoded port `8081`. If anything else is listening there the test errors with "address already in use". The existing part 4 already does this with 8080, so this isn't new, but for the new server it'd be nicer to bind to port 0 and read the assigned port back from the site/runner.

**Not caused by this PR, noting for a future follow-up**

`urllib.robotparser` applies the first matching rule, not the longest. With wired.com's real file (`Disallow: /*?` followed by `Allow: /*?page`), `/story/x/?page=2` is denied both before and after this change, on 3.10 and 3.13. RFC 9309 / Google use longest-match.

Thanks for the fix @Nalhin

Nalhin:
Great, thanks @SohamKukreti!

Can we get this landed?

Nalhin:
@SohamKukreti I've addressed the nits

ntohidi:
Merged — thanks @Nalhin! The `/*?` → `/*?*` rewrite is a neat fix: an empty query is the only thing `urlunparse` drops, so that one line covers the whole class of rules.

Digging into this turned up a second bug right next to it — the wildcard monkey patch breaks `Allow:` rules on Python 3.14, where the stdlib parser is already RFC 9309 compliant. Fixed separately in #2278.

Nalhin:
@ntohidi Thanks for getting this landed!

PR #2232, opened 2026-09-07
fix(docker): recycle pooled browser contexts by pages served (#2231)

### LINKED ISSUE #2231: [Bug]: Pooled browsers get ~5 s/navigation slower with use, and the janitor never recycles a busy pool
### crawl4ai version

0.9.2 (Docker server, unclecode/crawl4ai:0.9.2)

### Expected Behavior

A pooled browser serves the thousandth navigation about as fast as the first, or the pool replaces it when it stops doing so.

### Current Behavior

A pooled browser degrades with sustained use until it is roughly **5 s per navigation** slower than a browser created fresh in the same container, at the same instant, against the same host. Because `janitor()` only closes browsers that have been **idle** past a TTL, a server under continuous load never has an idle browser and therefore never recycles one. The degradation persists until the container restarts.

`GET /health` stays green throughout, so nothing an orchestrator watches reflects it.

Same workload, same code, same config, on one container 30 minutes apart:

| Sidecar state | wall clock, same job | pages |
|---|---|---|
| just restarted | **23.0 s** | 18/18 |
| ~1 h of crawling | **183.3 s** | 18/18 |

The same 20-URL batch, isolated:

```
degraded pool      128.2 s   (and 150.2 s -> HTTP 504 on the attempt before)
after restart       10.6 s / 10.4 s / 11.2 s
cold container      16.0 s   (first request after boot - cold start is not the problem)
```

The penalty is per-navigation and origin-independent. Measured in the same minute on the degraded container, comparing the pooled browser against a browser created fresh **inside that same container**:

```
                  degraded pooled browser     fresh browser
site A (Shopify)            6.59 s                2.12 s
example.com                 5.79 s                0.72 s
iana.org                    5.92 s                0.72 s
```

### What it is not - each ruled out by measurement

- **The target site.** `curl` TTFB 0.58 s; raw Playwright `goto` inside the container 0.56 s, while the pooled browser needed 6.59 s.
- **Target-side throttling of our IP.** A fresh browser was fast at the same instant, from the same container, to the same host.
- **DNS / egress.** `getaddrinfo` 0.02 s in the container.
- **Concurrency.** In-process `arun_many` of 12 URLs: 5.2 s. Three concurrent rounds back to back left single-page time at 1.0 s.
- **`context.route` handlers accumulating** from the `block_resources` hook (the factory calls `context.route("**/*")` per page on a shared context). A real suspect, but 30 pages with the hook attached stayed flat at 1.1 s.
- **Browser age or page count alone.** 60 sequential real pages on one crawler: flat 1.05-1.15 s. 160 inline `raw:` page creations: flat.

**I could not reproduce it in-process.** Four attempts failed to trigger it, so something about sustained *real* page crawling through the server's pooled crawler is responsible and I cannot name it. The threshold is somewhere between 60 pages (clean) and several hundred (badly degraded); I have not bracketed it further. I would rather report the effect with the evidence than guess at the mechanism.

### Why it never recovers on its own

`crawler_pool.janitor()` closes a browser only after it has been idle past `cold_ttl` / `hot_ttl`, and skips any crawler with `active_requests > 0`. Under steady load neither condition is ever met, so **the workload that causes the degradation is exactly the workload that prevents recycling.** `release_crawler` is called in a `finally`, so the count does return to zero - idleness, not leakage, is the blocker.

### Two suggestions

**1. Recycle on use, not only on idleness.** `USAGE_COUNT` and `LAST_USED` already exist in `crawler_pool.py`, so a `pool.max_uses` / `pool.max_age_s` sweep in `janitor()` is a small change: close a browser past either threshold when `active_requests == 0`, and let `get_crawler()` re-create it on demand. That is a mitigation, not a fix - it bounds the damage without explaining the cause. Happy to open a PR if you want it.

We run this today from outside the process, on a cron, against `/monitor/browsers` + `POST /monitor/actions/kill_browser`, and it works: after a kill the next `/crawl` completed in 569 ms. Doing it in the janitor would remove the need for anyone to build that.

Two things that do **not** work as recycling mechanisms, both measured:

- `POST /monitor/actions/restart_browser {"sig": "permanent"}` wedges the worker - it deadlocks the pool lock (#2230).
- gunicorn `--max-requests`. It fires correctly and then hangs in `Waiting for application shutdown.` because the ASGI shutdown blocks closing the browser pool; with the image's `--timeout 1800` the arbiter will not kill the stuck worker for 30 minutes. Every worker wedged within 7 minutes and the service returned 502 throughout.

**2. Make the pool's state visible to `/health`.** Every failure mode here - this one, a dead pooled browser after a `Page.goto` timeout (#842, PR #1946), and the restart deadlock - is invisible to `GET /health`, because health never touches the pool. An optional deep/readiness check would let an orchestrator restart a container that is answering nothing, instead of a human noticing a latency graph an hour later.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash

```

### Steps to Reproduce

```bash
1. Run the 0.9.2 Docker server with a pool config that keeps one browser hot.
2. Drive real (not `raw:`) page crawls through `/crawl` for ~30-60 minutes, several hundred pages, never letting the pool go idle past the TTL.
3. Time a single navigation through the pool, then create a browser directly inside the same container (`playwright` via `docker exec`) and time the same URL. The pooled one is ~5 s slower.
4. Restart the container and repeat step 3 - the gap is gone.
```

### Code snippets

```python

```

### OS

Linux (Docker)

### Python version

3.12 (image)

### Browser

Chromium (bundled)

### Browser version

_No response_

The gunicorn `--max-requests` attempt at recycling, which is how the shutdown-blocks-on-the-pool behaviour surfaced:

```
18:16:59 Maximum request limit of 63 exceeded. Terminating process.
18:16:59 Shutting down
18:16:59 Waiting for application shutdown.   <- stops here
```

`/health` latency went 1.2 ms -> 8 s and the proxy answered 502 for everything until the containers were restarted.

ntohidi:
Fixes #2231.

## The problem

A pooled browser in the Docker server gets slower with sustained use, and `janitor()` never recycles it — it only closes browsers that have been **idle** past a TTL, and a server under continuous load never has an idle one. The workload that causes the degradation is exactly the workload that prevents recycling.

## Where the slowdown actually lives

Measured, because the reporter could not reproduce it in-process and neither could I at first. After 1000 real page loads on one pooled browser, timing the same navigation three ways:

| | median navigation |
|---|---|
| the worn pooled context | **32.0 ms** |
| a new context, **same** Chromium process | **7.1 ms** |
| a brand-new Chromium process | **7.0 ms** |
| the worn context after clearing its state in place | **16.3 ms** |

A fresh context inside the same process is as fast as a fresh process, so the browser process is not the problem — the **context** is. A control run with pages that leave no state behind stayed flat over 2000 pages; the same pages with cookies + localStorage + a service worker degraded 4-5x from ~500 pages on.

That matches the code: `browser_manager.py` keeps one context per config for the life of the pooled browser, makes a new page per crawl, and never clears that context's cookies or storage. It also explains why the effect needs *real* pages — `raw:` pages leave no state.

## The change

`browser_manager` already has version-based recycling that replaces the context under load without waiting for a quiet moment (no idleness required, old contexts drain naturally). It was simply never enabled for the Docker server. This turns it on:

- `config.yml`: `crawler.pool.max_pages_before_recycle: 200`, next to the other pool policy (`max_pages`, `idle_ttl_sec`). `0` disables.
- `crawler_pool.py`: applied in `get_crawler()` / `init_permanent()`.

It is applied in the pool rather than in `crawler.browser.kwargs` because `/crawl` and `/crawl/stream` build their `BrowserConfig` from the request body — the server's browser kwargs never reach them. Every endpoint goes through `get_crawler()`, and pooling is what makes a browser long-lived, so the pool is the right owner of the policy.

Applied before `_sig()`, so all requests still share one signature and pooling is unchanged. An explicit per-request value wins.

No new lifetime logic in `janitor()` — that would put the same rule in two places.

## Effect

| pages | penalty off | penalty with recycling at 200 |
|---|---|---|
| 500 | 31 ms | 10 ms |
| 1500 | 29 ms | 14 ms |

It **bounds** the rot to one recycle window rather than removing it. The threshold is the knob.

## Caveats, stated plainly

- My reproduction shows a 25 ms penalty, not the ~5 s the reporter measured. Same shape, much smaller — local pages are cheap, and I wore out one origin's state, not hundreds.
- My penalty was same-origin only; the reporter's was origin-independent. That gap is not closed, so this is "the right lever", not "proven sufficient at their scale". @talelboussetta, if you can run this against your workload, that is the confirmation this needs.
- The `/health` suggestion in the issue is not addressed here; it is a separate concern.

## Tests

`tests/docker/test_pool_recycle_config.py` — covers the config value, the pool applying it to a request-supplied config, an explicit caller value winning, signature stability, and the `get_crawler()` call site. Removing the config line fails 2 tests; removing the call site fails 1.

https://claude.ai/code/session_019EQGr8ey8riKFr4inp9B8f

### DIFF
--- a/deploy/docker/config.yml
+++ b/deploy/docker/config.yml
@@ -88,6 +88,14 @@ crawler:
     kwargs:
       headless: true
       text_mode: true
+      # Pooled browsers are long-lived, and their browser context is reused for
+      # every crawl that shares a config. Real sites leave state in that context
+      # (cookies, localStorage, service workers) which is never cleared, and
+      # navigation gets measurably slower as it builds up. Recycling by pages
+      # served bounds that: the context is replaced inside the same Chromium
+      # process, old contexts drain on their own, and it works under sustained
+      # load (unlike the janitor, which only closes *idle* browsers). See #2231.
+      max_pages_before_recycle: 200
     extra_args:
       # - "--single-process"
       # SECURITY: --no-sandbox disables the Chromium renderer sandbox (a renderer
--- a/deploy/docker/config.yml
+++ b/deploy/docker/config.yml
@@ -84,18 +84,18 @@ crawler:
   pool:
     max_pages: 40                          # ← GLOBAL_SEM permits
     idle_ttl_sec: 300                     # ← 30 min janitor cutoff
+    # Pooled browsers are long-lived and reuse one browser context per config.
+    # Real sites leave state in that context (cookies, localStorage, service
+    # workers) which is never cleared, and navigation gets measurably slower as
+    # it builds up. Recycling by pages served bounds it: the context is replaced
+    # inside the same Chromium process, old contexts drain on their own, and it
+    # works under sustained load — unlike the janitor, which only closes *idle*
+    # browsers and so never fires on a busy server. 0 disables. See #2231.
+    max_pages_before_recycle: 200
   browser:
     kwargs:
       headless: true
       text_mode: true
-      # Pooled browsers are long-lived, and their browser context is reused for
-      # every crawl that shares a config. Real sites leave state in that context
-      # (cookies, localStorage, service workers) which is never cleared, and
-      # navigation gets measurably slower as it builds up. Recycling by pages
-      # served bounds that: the context is replaced inside the same Chromium
-      # process, old contexts drain on their own, and it works under sustained
-      # load (unlike the janitor, which only closes *idle* browsers). See #2231.
-      max_pages_before_recycle: 200
     extra_args:
       # - "--single-process"
       # SECURITY: --no-sandbox disables the Chromium renderer sandbox (a renderer
--- a/deploy/docker/crawler_pool.py
+++ b/deploy/docker/crawler_pool.py
@@ -20,7 +20,24 @@ LOCK = asyncio.Lock()
 # Config
 MEM_LIMIT = CONFIG.get("crawler", {}).get("memory_threshold_percent", 95.0)
 BASE_IDLE_TTL = CONFIG.get("crawler", {}).get("pool", {}).get("idle_ttl_sec", 300)
+RECYCLE_PAGES = CONFIG.get("crawler", {}).get("pool", {}).get("max_pages_before_recycle", 0)
 DEFAULT_CONFIG_SIG = None  # Cached sig for default config
+
+
+def _apply_pool_defaults(cfg: BrowserConfig) -> BrowserConfig:
+    """Apply pool-owned policy to a browser config before it is pooled.
+
+    Browsers handed out by this pool are long-lived, and the endpoints build
+    their BrowserConfig from the request body (see api.py handle_crawl_request),
+    so config.yml's browser kwargs never reach them. Pooling is what makes a
+    browser long-lived, so the pool is where its recycling policy belongs.
+
+    Applied before _sig() so every request shares one signature and pooling is
+    unaffected. An explicit per-request value wins. See #2231.
+    """
+    if RECYCLE_PAGES and not cfg.max_pages_before_recycle:
+        cfg.max_pages_before_recycle = RECYCLE_PAGES
+    return cfg
 
 
 def get_pool_snapshot() -> dict:
@@ -54,7 +71,7 @@ def _is_default_config(sig: str) -> bool:
 
 async def get_crawler(cfg: BrowserConfig) -> AsyncWebCrawler:
     """Get crawler from pool with tiered strategy."""
-    sig = _sig(cfg)
+    sig = _sig(_apply_pool_defaults(cfg))
     async with LOCK:
         # Check permanent browser for default config
         if PERMANENT and _is_default_config(sig):
@@ -135,7 +152,7 @@ async def init_permanent(cfg: BrowserConfig):
     async with LOCK:
         if PERMANENT:
             return
-        DEFAULT_CONFIG_SIG = _sig(cfg)
+        DEFAULT_CONFIG_SIG = _sig(_apply_pool_defaults(cfg))
         logger.info("🔥 Creating permanent default browser")
         PERMANENT = AsyncWebCrawler(config=cfg, thread_safe=False)
         await PERMANENT.start()
--- a/tests/docker/test_pool_recycle_config.py
+++ b/tests/docker/test_pool_recycle_config.py
@@ -1,0 +1,42 @@
+"""The Docker pool must recycle its browser context by pages served (#2231).
+
+The pool's janitor only closes *idle* browsers, so a server under sustained
+load never recycles one. Recycling by page count is what bounds the context
+state (cookies/localStorage/service workers) that slows navigation down.
+"""
+import sys
+from pathlib import Path
+
+import pytest
+import yaml
+
+from crawl4ai import BrowserConfig
+
+CONFIG_PATH = Path(__file__).resolve().parents[2] / "deploy" / "docker" / "config.yml"
+
+
+@pytest.fixture(scope="module")
+def browser_section():
+    return yaml.safe_load(CONFIG_PATH.read_text())["crawler"]["browser"]
+
+
+def test_recycle_is_enabled(browser_section):
+    kwargs = browser_section.get("kwargs", {})
+    assert kwargs.get("max_pages_before_recycle", 0) > 0, (
+        "deploy/docker/config.yml must set crawler.browser.kwargs."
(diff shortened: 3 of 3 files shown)


talelboussetta:
Thanks for digging into this — the context-vs-process split is a much better answer than anything I got to, and the fresh-context-in-the-same-process measurement is the part that settles it.

On the gap you flagged between your ~25 ms and my ~5 s: I think that is scale in the same mechanism rather than two different effects. My workload crawls several hundred distinct origins through one pooled context; you wore out one origin. Cookie jar entries, localStorage, service worker registrations and CacheStorage all accumulate per origin inside that single context, so the number of origins is a multiplier on exactly the state you identified.

It also fits the origin-independence I saw. My control was a fresh browser in the same container hitting example.com and iana.org while the worn pooled one was slow to both. If the cost lived in the visited origin's own state, the worn browser would have been fast on origins it had never touched, and it was not. A context-wide cost that every navigation pays is consistent with both sets of numbers. Offered as an explanation that fits the data, not something I have proven.

Two review notes.

**The new test file errors at collection when the deps are absent, rather than skipping.** `from crawl4ai import BrowserConfig` is a top-level import that runs before `pytest.importorskip("crawler_pool")`, so the guard never gets its chance — and your skip reason suggests that case was meant to be handled. On a checkout without the server deps installed, collection dies in `crawl4ai/ssl_certificate.py` on `import OpenSSL.crypto` with `ModuleNotFoundError: No module named 'OpenSSL'`, giving `1 error during collection` instead of a skip. Pulling `BrowserConfig` in through an importorskip as well fixes it.

**`max_pages_before_recycle` is on `UNTRUSTED_FIELD_ALLOWLIST["BrowserConfig"]`**, so "an explicit per-request value wins" is reachable from the wire and not only from `config.yml`. Two consequences that might deserve a sentence in the config comment: an operator can A/B the knob per request against a stock image with no redeploy, and a caller sending a value that differs from the pool default forks `_sig()` into a second pooled browser.

On confirmation at my scale: yes, I will run it and post numbers back here. The allowlist point above makes that cheaper than I expected, since I can set the field per request against the pinned image I already run and keep the recycling behaviour as the only thing under test. I will report whatever I get, including if it fails to reproduce your improvement.

talelboussetta:
Ran the knob against a local rig. Three results, one of them a null.

**1. The field applies from an untrusted request body on stock 0.9.2.** Sending `max_pages_before_recycle: 200` inside `browser_config.params` to an unmodified `unclecode/crawl4ai:0.9.2` produced, in the server log:

`[BROWSER] ℹ Bumping browser version 1 -> 2 after 200 pages (1 active, 0 idle sigs)`

and 15 more across a 3000-page run. That is worth a line in the PR: operators can adopt this today without waiting for a release, and it makes the knob A/B-able against an image they already run. I checked the log rather than the status code, since an unrecognised key is dropped silently with a 200.

**2. Recycling is free.** 3000 pages per arm, fresh container per arm, `browser_config` identical apart from the field. The probe is a single navigation to a dedicated origin, timed every 250 pages.

| arm | median probe | spread | drift, first half to second half |
|---|---|---|---|
| control, no recycling | 337.5 ms | 63.7 ms | +1.9 ms |
| recycle at 200, 15 recycles fired | 319.1 ms | 50.4 ms | -2.4 ms |

15 context replacements cost nothing measurable — the recycling arm was, if anything, marginally faster. That seems like the result you would want in hand before defaulting it on, and it supports doing so.

**3. I did not reproduce the decay, and my rig cannot refute your 25 ms.** 3000 page loads across 99 distinct origins, each setting 5 cookies, ~20 KB of localStorage and an IndexedDB store, left the probe flat: +1.9 ms of drift.

Two reasons not to read that as a negative result:

- Resolution. My probe times a whole `/crawl` round trip, ~320 ms with a 50-60 ms spread, so a 25 ms navigation delta sits inside my noise. The rig can see a multi-second effect; it cannot see yours.
- A missing ingredient. No service workers. The fixture is plain HTTP, which is not a secure context, so `serviceWorker.register()` never runs. Service workers are one of the three state types you named, and cookies plus localStorage plus IndexedDB alone did not do it at this scale.

So the honest reading is that my synthetic rig failed to reproduce the mechanism, not that the mechanism is absent.

The real-workload confirmation is still what I owe you, and I have not run it: the deployment that shows the effect is in production and I am not going to experiment on it. I will get it onto a staging copy and post numbers. If it is useful sooner, I can rebuild the local rig over HTTPS so service workers actually register, which would test your third ingredient directly — say the word and I will run that next.

PR #2244, opened 2026-09-09
Fix: O(n²) parent lookup in BFS and duplicate URL enqueue in BestFirst

### LINKED ISSUE #2242: [Bug]: BFSDeepCrawlStrategy re-scans the full level for parent lookup (O(n²)); BestFirstCrawlingStrategy can enqueue the same URL twice
### crawl4ai version

0.9.3

### Expected Behavior

- Per-level bookkeeping in BFSDeepCrawlStrategy (matching a fetched result back to its parent URL) should scale linearly with the number of URLs in a level.
- A URL discovered by multiple pages in BestFirstCrawlingStrategy should be scored and queued once, not once per inbound link.

### Current Behavior

1. **O(n²) parent lookup** - `deep_crawling/bfs_strategy.py`, `_arun_batch` and `_arun_stream`:

`parent_url = next((parent for (u, parent) in current_level if u == url), None)`

This linearly re-scans the entire `current_level` list once per fetched result. For a level of N URLs this is O(N²) pure-Python work with no I/O involved - a page with high fan-out (hundreds/thousands of links) makes this bookkeeping step dominate.

2. **Duplicate enqueue** - `deep_crawling/bff_strategy.py`, `link_discovery`/ `_arun_best_first`:
`visited` is only populated when an item is dequeued, not when it's discovered, so `link_discovery`'s dedup check `(if base_url in visited: continue`) doesn't stop the same URL being pushed onto the priority queue twice if two different pages link to it before it's first processed. The duplicate is silently dropped later at dequeue time, so output is still correct, but it was scored and queued for nothing.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
- URL(s): Any site with high link fan-out per page, or shared links between sibling pages
- Settings used: BFSDeepCrawlStrategy(max_depth=2, max_pages=1000+) for issue 1; BestFirstCrawlingStrategy(...) with a url_scorer for issue 2
```

### Steps to Reproduce

```bash
1. Run `BFSDeepCrawlStrategy` against a page with hundreds+ of internal links at depth 1; profile `_arun_batch`'s per-result loop - the parent lookup dominates as level size grows.
2. Run `BestFirstCrawlingStrategy` against a small site where ≥2 pages link to the same third page before it's first crawled; log `queue.qsize()` or inspect enqueued items - the shared URL appears twice.
```

### Code snippets

```python

```

### OS

macOS 26.6.2 (Darwin 25.6.0)

### Python version

3.13.12

### Browser

_No response_

### Browser version

_No response_

_No response_

yashikam19:
Fixes #2242

## Summary
Fixes two related inefficiencies in the deep-crawl strategies found while
integrating `BestFirstCrawlingStrategy` for site-wide crawling:

1. `BFSDeepCrawlStrategy._arun_batch`/`_arun_stream` re-scanned the entire
   `current_level` list to find each result's parent URL - O(n²) for a level
   of size n. Replaced with a dict built once per level (`dict(current_level)`),
   O(n).
2. `BestFirstCrawlingStrategy` only tracked `visited` (updated at dequeue
   time), so a URL discovered via multiple parent pages before being crawled
   could be pushed onto the priority queue more than once - wasted scorer
   calls and queue slots. Added a `queued` set updated at enqueue time to
   prevent this.

Both are internal-only fixes - no behavior or public API change, same outputs.

## Files Changed
- `crawl4ai/deep_crawling/bfs_strategy.py` - build `parent_by_url` dict once
  per level in `_arun_batch` and `_arun_stream`, instead of a linear scan per
  result.
- `crawl4ai/deep_crawling/bff_strategy.py` - added a `queued: Set[str]` tracked
  alongside `visited`, checked/updated at the point a URL is pushed onto the
  priority queue (both fresh-start and resume-from-state paths).

## How Has This Been Tested?
- `pytest tests/deep_crawling/ tests/general/test_deep_crawl*.py`: 70 passed,
  0 regressions 
- Isolated benchmark on real production data: pulled djangoproject.com's
  actual `sitemap.xml` (1000 URLs, one legitimate request, no crawling) and
  timed the parent-lookup code directly - 11.69ms (old) → 0.08ms (new), ~145x.

## Checklist:

### DIFF
--- a/crawl4ai/deep_crawling/bff_strategy.py
+++ b/crawl4ai/deep_crawling/bff_strategy.py
@@ -216,6 +216,8 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
             queue_items = self._resume_state.get("queue_items", [])
             for item in queue_items:
                 await queue.put((item["score"], item["depth"], item["url"], item["parent_url"]))
+            # `queued` tracks every URL ever pushed onto the queue
+            queued: Set[str] = set(visited) | {item["url"] for item in queue_items}
             # Initialize shadow list if callback is set
             if self._on_state_change:
                 self._queue_shadow = [
@@ -227,6 +229,7 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
             initial_score = self.url_scorer.score(start_url) if self.url_scorer else 0
             await queue.put((-initial_score, 0, start_url, None))
             visited: Set[str] = set()
+            queued: Set[str] = {start_url}
             depths: Dict[str, int] = {start_url: 0}
             # Initialize shadow list if callback is set
             if self._on_state_change:
@@ -313,6 +316,9 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
                     await self.link_discovery(result, url, depth, visited, new_links, depths)
                     
                     for new_url, new_parent in new_links:
+                        # Skip URLs already sitting in the queue
+                        if new_url in queued:
+                            continue
                         new_depth = depths.get(new_url, depth + 1)
                         new_score = self.url_scorer.score(new_url) if self.url_scorer else 0
                         # Skip URLs with scores below the threshold
@@ -322,6 +328,7 @@ class BestFirstCrawlingStrategy(DeepCrawlStrategy):
                             )
                             self.stats.urls_skipped += 1
                             continue
+                        queued.add(new_url)
                         queue_item = (-new_score, new_depth, new_url, new_parent)
                         await queue.put(queue_item)
                         # Add to shadow list if tracking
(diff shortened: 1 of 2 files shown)


ntohidi:
Thanks for this, and sorry for the slow turnaround. Both bugs you spotted are real and the diagnosis was right. We've landed the fix in #2265 (merged to `develop` as `2740268`), so I'm closing this one as superseded.

The BFS half there is essentially your change. On the BestFirst half we went a different way: de-duplicating on identity — whether the URL has been seen at all — turns out to drop pages. Best-first doesn't visit levels in order, so a URL can be discovered at depth 3 through a high-scoring branch and only later at depth 2 through a slower one. The duplicate enqueue your PR removes was quietly doing a second job: the priority queue popped the shallower copy first, so a URL ended up at its true distance from the start. Skipping the re-enqueue freezes the deeper depth, and then `link_discovery` bails at `4 > max_depth` and that whole subtree goes missing.

#2265 keeps your de-duplication but keys it on depth instead — a URL already queued at an equal or shallower depth is skipped, a strictly shallower rediscovery still re-queues. Same win you were after, without the lost pages.

Genuinely useful report — the O(n²) lookup was costing ~46% of runtime on a wide level. Thanks for digging into it.

PR #2250, opened 2026-09-11
docs: fix dead micronaut guide URL in README example

jrlprost:
The heuristic markdown generation example in README.md crawls `https://docs.micronaut.io/4.9.9/guide/`, which now returns a 404 (the Micronaut docs no longer serve that pinned version path).

Verified with `curl -sI https://docs.micronaut.io/4.9.9/guide/` (404) vs `curl -sI https://docs.micronaut.io/latest/guide/` (200), and updated the example to use the `latest` path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,7 +393,7 @@ async def main():
     
     async with AsyncWebCrawler(config=browser_config) as crawler:
         result = await crawler.arun(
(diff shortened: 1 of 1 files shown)


Ryan-cw:
useful

jrlprost:
Thanks for confirming!

PR #2253, opened 2026-09-12
fix(robots): verify TLS when fetching robots.txt

### LINKED ISSUE #2252: [security] RobotsParser.can_fetch() disables TLS verification (ssl=False) when fetching robots.txt
## Description

In `crawl4ai/utils.py` (line 344), `RobotsParser.can_fetch()` issues the robots.txt request with `ssl=False`:

```python
session.get(robots_url, timeout=2, ssl=False)
```

Certificate validation is disabled, so an on-path attacker (MITM, rogue DNS, malicious proxy) can impersonate the target host and feed arbitrary robots.txt content, silently controlling whether the crawler honors `Disallow` rules for the target site.

Impact is limited (robots.txt is an advisory protocol and the request carries no credentials), so I'm filing this as a hardening issue rather than a private security report.

## Suggested fix

Keep TLS verification enabled for `https://` URLs; allow opt-in disabling only for explicitly configured hosts:

```python
ssl = parsed.scheme.lower() != "https"  # or an explicit per-request override
```

## Reproduction

1. Route the target host through a proxy that presents a self-signed certificate.
2. Observe the robots.txt request succeeding despite the invalid certificate, and the tampered `Disallow` rules taking effect.

zhongxiao-chang:
## Problem
RobotsParser.can_fetch() fetched robots.txt with ssl=False, disabling TLS certificate
verification for every https target (crawl4ai/utils.py). An on-path attacker (MITM /
rogue DNS / malicious proxy) could impersonate the host, forge robots.txt and silently
control whether the crawler honours Disallow rules. Reported in #2252.

## Fix
- Keep certificate verification by default: pass ssl=None (aiohttp default context)
  instead of the unconditional ssl=False.
- Preserve an explicit escape hatch for self-hosted mirrors / debugging:
  RobotsParser(verify_ssl=False).
- Plain http:// behaviour is unchanged (aiohttp ignores ssl there).

## Tests
- New offline regression test tests/test_robots_tls_verification.py
  (aiohttp.ClientSession is stubbed; no network):
  * default https fetch keeps verification (ssl is None, not False)
  * verify_ssl=False still disables it (escape hatch)
  * Disallow rules still enforced; 404/connection errors still allow
- pytest tests/test_robots_tls_verification.py -q  ->  4 passed

Closes #2252

### DIFF
--- a/crawl4ai/utils.py
+++ b/crawl4ai/utils.py
@@ -253,9 +253,13 @@ class RobotsParser:
     # Default 7 days cache TTL
     CACHE_TTL = 7 * 24 * 60 * 60
 
-    def __init__(self, cache_dir=None, cache_ttl=None):
+    def __init__(self, cache_dir=None, cache_ttl=None, verify_ssl=True):
         self.cache_dir = cache_dir or os.path.join(get_home_folder(), ".crawl4ai", "robots")
         self.cache_ttl = cache_ttl or self.CACHE_TTL
+        # TLS verification is ON by default: an on-path attacker must not be able to
+        # forge robots.txt and silently flip Disallow rules (see issue #2252).
+        # Opt out explicitly (self-hosted mirrors / debug only) via verify_ssl=False.
+        self.verify_ssl = verify_ssl
         os.makedirs(self.cache_dir, exist_ok=True)
         self.db_path = os.path.join(self.cache_dir, "robots_cache.db")
         self._init_db()
@@ -341,7 +345,11 @@ class RobotsParser:
                 robots_url = f"{scheme}://{domain}/robots.txt"
                 
                 async with aiohttp.ClientSession() as session:
-                    async with session.get(robots_url, timeout=2, ssl=False) as response:
+                    # aiohttp's `ssl=None` means "use the default context" (certificates
+                    # verified). For plain http:// the argument is ignored, so this stays
+                    # a no-op for http while restoring verification for https (#2252).
+                    ssl_ctx = None if self.verify_ssl else False
+                    async with session.get(robots_url, timeout=2, ssl=ssl_ctx) as response:
                         if response.status == 200:
                             rules = await response.text()
                             self._cache_rules(domain, rules)
--- a/tests/test_robots_tls_verification.py
+++ b/tests/test_robots_tls_verification.py
@@ -1,0 +1,106 @@
+"""Offline regression tests for robots.txt TLS verification (issue #2252).
+
+No network access: aiohttp.ClientSession is replaced by a recorder that exposes the
+kwargs the parser passes to ``session.get``.
+"""
+
+import asyncio
+
+from crawl4ai import utils
+from crawl4ai.utils import RobotsParser
+
+
+class _FakeResponse:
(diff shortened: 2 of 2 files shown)


liuchunyi-buaa:
Reviewed the diff — this matches the fix we had in mind:

1. **Secure by default**: passing `ssl=None` keeps aiohttp's verified default context; `ssl=False` is no longer used implicitly.
2. **Explicit escape hatch**: the per-instance `verify_ssl=False` opt-out is a reasonable shape for the rare legitimate case (self-hosted mirrors/debugging), and it's backward compatible since the new parameter defaults to `True`.
3. **No change for plain `http://`** — aiohttp ignores `ssl` there, as noted in the comment.
4. **Tests**: the offline regression tests cover both the `ssl` kwarg semantics (None vs False) and the unchanged Disallow/404 behaviour — nice, no network dependency.

LGTM — thanks for the clean fix and tests!

zhongxiao-chang:
Thanks for the review — glad the diff matches what you had in mind (and yes,
kept on the public path as you noted, nothing handled privately).

One thing worth flagging while this sits in the queue: CI never ran on this PR,
so here is the local evidence in its place.

Why nothing ran:
1) The only workflow run against head 7f56ae5 is "Discord GitHub Notifications"
   and it sits at conclusion=action_required (first-time fork contributor,
   waiting for maintainer approval) — a notification job, not a test job.
2) security.yml is the only active workflow here with a pull_request trigger,
   and its paths filter is deploy/docker/**, crawl4ai/async_configs.py,
   .github/workflows/security.yml. This PR touches crawl4ai/utils.py plus one
   new test file, so it falls outside that filter.
3) No other active workflow triggers on pull_request (main.yml = notifications,
   release.yml / docker-release.yml = tags, test-release.yml is disabled).

So check-runs = 0 means no test workflow was triggered — not that anything failed.

Local run:
    $ python -m pytest tests/test_robots_tls_verification.py -q
    ....  [100%]  4 passed in 5.75s
Offline and deterministic (session stubbed; no network / Redis / Docker). Covers
ssl=None (default verifying context) vs verify_ssl=False (explicit opt-out),
plus unchanged plain-http and Disallow/404 behaviour.

Diff unchanged: crawl4ai/utils.py + tests/test_robots_tls_verification.py
(+116/-2, 2 files).
No action needed — the branch is stable and can be rebased on develop on request.

Ricchkdbanks:
,l

ntohidi:
Thanks for the contribution. However, I will not merge it since the fix is already done on our side for the next release

zhongxiao-chang:
Understood - thanks for the clarification. Noted on SECURITY.md: I will route any future security reports to unclecode@crawl4ai.com / nasrin@crawl4ai.com rather than a public issue, and I will leave the PR closed.

One thing that may still be useful before your release: this PR ships an offline regression test (`tests/test_robots_tls_verification.py` - stubbed session, no network) covering the `ssl=None` default-verification path and the explicit `verify_ssl=False` opt-out. Happy to post it here as a plain file, or open a small test-only PR if that is easier - your call, no obligation either way.

Thanks for the quick follow-up.

PR #2259, opened 2026-09-13
Fix: DefaultTableExtraction drops <th> row headers and ignores rowspan

### LINKED ISSUE #2258: [Bug]: DefaultTableExtraction drops <th> row headers and ignores rowspan, producing misaligned result.tables
### crawl4ai version

0.9.3

### Expected Behavior

`result.tables[i]["rows"]` should match the logical grid a browser renders:
- Row-header cells (`<th scope="row">` inside `<tbody>`) are kept as the first column.
- A `rowspan` cell's value is repeated in every row it covers.
- cells keep the current "repeat the value" behaviour, but the row must stay aligned with the header (no left shift, no phantom trailing "").

### Current Behavior

`DefaultTableExtraction.extract_table_data` only iterates `.//td` for body rows, so every `<th>` inside `<tbody>` is dropped: the remaining cells shift left and the last column is padded with "". `rowspan` is not handled at all. For manual documentation tables whose first column is a product / parameter name in `<th>`, the whole key column disappears from `result.tables`.

This is the follow-up to #2007: since v0.9.1 `rowspan`/`colspan` survive in `cleaned_html`, but the extractor does not use them.

### Is this reproducible?

Yes

### Inputs Causing the Bug

```bash
Any table with (a) `<th scope="row">` cells in the body, or (b) a `rowspan` > 1 cell. Minimal HTML is in the steps below.
```

### Steps to Reproduce

```bash
1. Run the script below (no browser, no crawl).
2. Compare the printed rows with the expected rows.
```

### Code snippets

```python
from lxml import html as lhtml
from crawl4ai import DefaultTableExtraction

HTML = """<table>
<tr><th></th><th>Feature A</th><th>Feature B</th></tr>
<tr><th scope="row">Item 1</th><td>yes</td><td>yes</td></tr>
<tr><th scope="row">Item 2</th><td>no</td><td>yes</td></tr>
</table>
<table>
<thead><tr><th>Group</th><th>Option X</th><th>Option Y</th></tr></thead>
<tbody>
<tr><td rowspan="2">Group 1</td><td>value x</td><td>value y</td></tr>
<tr><td colspan="2">note that applies to X and Y</td></tr>
</tbody></table>"""

root = lhtml.fromstring(HTML)
for table in DefaultTableExtraction().extract_tables(root):
    print(table["headers"], table["rows"])
```

### OS

windows 11

### Python version

3.12

### Browser

N/A

### Browser version

N/A

Actual:
['', 'Feature A', 'Feature B'] [['yes', 'yes', ''], ['no', 'yes', '']]
['Group', 'Option X', 'Option Y'] [['Group 1', 'value x', 'value y'], ['note that applies to X and Y', 'note that applies to X and Y', '']]

Expected:
['', 'Feature A', 'Feature B'] [['Item 1', 'yes', 'yes'], ['Item 2', 'no', 'yes']]
['Group', 'Option X', 'Option Y'] [['Group 1', 'value x', 'value y'], ['Group 1', 'note that applies to X and Y', 'note that applies to X and Y']]

I'd like to submit a PR: expand rowspan/colspan into a rectangular grid (pending-cell map), collect `th|td` for body rows, and add unit tests for the cases above. The change is limited to `crawl4ai/table_extraction.py` plus a small shared grid helper; the scoring logic in `is_data_table` is untouched.

ylcoldplayer:
## Summary
- `DefaultTableExtraction.extract_table_data` only iterated `.//td` for body rows, so a row-header `<th>` inside `<tbody>` was dropped: every following cell shifted left and the row was padded with a phantom trailing `""`.
- `rowspan` was ignored entirely, so a spanning cell's value was not repeated into the rows it covers, misaligning `result.tables[i]["rows"]` against the header.
- Body rows are now walked left-to-right, expanding `colspan`/`rowspan` into a rectangular grid via a per-column "pending rowspan" map, collecting `th|td` cells (direct children, in document order). The implicit "first row is the header" case (no `<thead>`) is tracked explicitly so that row isn't double-counted now that `<th>` cells are also collected for body rows.
- Scoring logic in `is_data_table` is untouched.

Fixes #2258.

## Test plan
- [x] Added `tests/test_issue_2258_table_rowspan.py` covering: `<th scope="row">` row headers in `<tbody>`, a `rowspan=2` cell whose value must repeat into the next row, a `rowspan=3` cell repeating across all three covered rows, and a plain table without spans (regression guard).
- [x] Verified against the exact repro HTML from #2258 — output now matches the issue's expected `headers`/`rows`.
- [x] `pytest tests/test_issue_2258_table_rowspan.py tests/test_eval_security_adversarial.py -q` passes (one pre-existing unrelated failure in the adversarial suite — a missing `deploy/docker/hook_manager.py` fixture path in this sandboxed checkout — confirmed unrelated to this change).

https://claude.ai/code/session_01FmBabqc17vZKRUXfZzoUXY


ntohidi:
Thanks for this — you found the bug and the diagnosis was right on both counts.

We went with #2261 instead, for one reason: the `pending_rowspans` map survives a short row. Given

```html
<tr><td>a</td><td>b</td><td rowspan="2">SPAN</td></tr>
<tr><td>c</td></tr>
<tr><td>x</td><td>y</td><td>z</td></tr>
```

row 3 comes out `['x','y','SPAN']` — `z` is overwritten. Reproduced on your branch. #2261 writes spans forward into the grid instead, so no state carries between rows.

Your `<th scope="row">` fix and the `implicit_header_row` guard were the right calls and both landed in #2261. Closing this one — thanks for the report and the fix.

PR #2265, opened 2026-09-14
fix(deep-crawl): drop the O(n²) parent scan and the duplicate best-first enqueue

### LINKED ISSUE #2242: [Bug]: BFSDeepCrawlStrategy re-scans the full level for parent lookup (O(n²)); BestFirstCrawlingStrategy can enqueue the same URL twice

ntohidi:
Fixes #2242.

## The two bugs

**1. BFS re-scanned the whole level to find each result's parent** — `bfs_strategy.py`, both `_arun_batch` and `_arun_stream`:

```python
parent_url = next((parent for (u, parent) in current_level if u == url), None)
```

`current_level` is already a list of `(url, parent)` pairs — a dict spelled as a list. Building it once per level next to the existing `urls` line makes this O(n) instead of O(n²). `reversed()` preserves the old first-parent-wins result if a resumed level repeats a URL.

**2. BestFirst could queue the same URL twice** — root cause is that `bff_strategy.py` marked a URL `visited` at *dequeue* time, so between discovery and dequeue several parents could each push it. BFS and DFS both already mark at *discovery* time; BFF was the odd one out. Fixed by matching them, which makes the dequeue guard unreachable — so it goes, rather than adding a second set to track the same fact.

Restoring a checkpoint now de-dupes the saved queue and folds queued URLs into `visited`, so states written by older builds (which recorded only crawled URLs and could hold a repeat) still resume without a duplicate crawl.

Net: 8 lines added, 5 removed.

## Reproduced first

The offending lookup, in isolation:

| level size | time | growth per doubling |
|---|---|---|
| 1000 | 11.5 ms | — |
| 2000 | 38.3 ms | 3.33x |
| 4000 | 158.2 ms | 4.13x |
| 8000 | 667.3 ms | 4.22x |

In a full mock crawl of an 8000-link level with no network at all, that one line was ~46% of total runtime. After: 1439 ms → 741 ms, growth back to ~2.0x per doubling.

BestFirst, with root → a, b and both a and b linking to `shared`:

```
after page 2: queue=['/shared']
after page 3: queue=['/shared', '/shared']   <-- duplicate
```

## Tests

`tests/deep_crawling/test_deep_crawl_perf_2242.py`, 4 tests:

- `parent_url` correct for every child, batch and stream
- level bookkeeping scales linearly — uses `resume_state` + `max_depth=0` to isolate the lookup from link discovery, and a plain `FakeResult` because `MagicMock` creation swamped the signal
- BestFirst queues a shared URL once

All 4 fail on unpatched source and pass with the fix.

## Checking nothing else broke

- `link_discovery`, `_arun_best_first`, `_queue_shadow` and `state["visited"]` have no consumers outside `deep_crawling/`. `DFSDeepCrawlStrategy` subclasses BFS but overrides all three touched methods.
- End-to-end A/B on a local 41-page site with a dense shared-link graph, 5 strategy configs × batch and stream, through the real browser. BestFirst and DFS: identical URL sets, parents and depths. BFS differed — but running the *same* code twice differs by the same amount, so BFS result ordering is nondeterministic on its own. The order-independent invariant (every `parent_url` was itself crawled, `depth == parent depth + 1`) gives 24 violations before and 24 after: zero introduced. Those 24 come from a pre-existing BFS-batch bug where the root is crawled twice — visible on `develop` too, left alone here.
- Suite across `deep_crawling`, `deep_crwaling`, `general`, `regression`: 225 passed / 3 failed on `develop` → 229 passed / 3 failed with this branch (+4 = the new tests). Same three failures both sides (2 `ContentTypeFilter` cases, and `test_bff_scoring::test_best_first_strategy`, an unmarked `async def` pytest cannot collect — worth fixing separately).

Supersedes #2244, which fixes the same two bugs; the BFS half here is identical to that PR.

https://claude.ai/code/session_01Utx8BTDqXTscjquLV7vgLq


ntohidi:
Confirmed, and fixed in e5e1011. Good catch — the repro reproduces exactly as described, 18 pages on develop vs 17 on the PR, and X frozen at depth 3.

The part I missed: **the duplicate enqueue was load-bearing.** It was doubling as the depth relaxation. Best-first does not visit levels in order, so a URL can be found at depth 3 down a high-scoring branch and only later at depth 2 down a slower one. Develop's second enqueue carried the shallower depth and won the priority tie, so X settled at depth 2. De-duplicating on identity threw that away and took the subtree under LEAF with it.

So the fix now de-duplicates on **depth** rather than on identity — the whole change to `bff_strategy.py` is one guard in `link_discovery`:

```python
queued_depth = depths.get(base_url)
if queued_depth is not None and queued_depth <= new_depth:
    continue
```

A URL already queued at an equal or shallower depth is skipped — that is the waste #2242 reported, and it covers the reported case, where two same-depth parents link to one child. A strictly shallower re-discovery still re-queues, and the existing dequeue guard drops the stale deeper copy.

This also let me revert the riskier parts of the first cut: `visited` goes back to its original crawled-only meaning, the dequeue guard stays, and the resume-state compatibility shim is gone since nothing about checkpoint semantics changes any more. The BFS half is untouched.

Your repro is now a test, fillers and all, with your note about why they are there in the docstring: `test_best_first_keeps_the_shallowest_depth_for_a_shared_url`. It passes on develop, fails on the previous version of this PR, passes now — so the regression cannot come back silently.

Re-verified: 230 passed across `deep_crawling`/`deep_crwaling`/`general`/`regression` (3 failures pre-existing on develop), and the live A/B on the 41-page site is identical to develop for best-first and DFS including depths.

You were also right about why my A/B missed it — with `BATCH_SIZE=10` and no fillers, B and C land in the same batch and the ordering never arises. **#2244 has the same bug** and no test covering it.

